PR #2, opened 2018-04-14
Scaffold Staff Interface

indocomsoft:
Not sure whether I am doing it the right way, do give me feedback on how to correctly do this.


lib/cadet_web/plug/check_admin.ex
```diff
@@ -0,0 +1,22 @@
+defmodule CadetWeb.Plug.CheckAdmin do
+  @moduledoc """
+  A plug that checks whether :current_user is an admin
```
evansb: `A plug` is redundant, we know it's a Plug from the module name

evansb: A more detailed (but not too verbose) explanation will do, e.g what happens if user is not admin?

lib/cadet_web/plug/check_admin.ex
```diff
@@ -0,0 +1,22 @@
...
+  """
+
+  import Plug.Conn
+
+  import Phoenix.Controller
+
+  def init(opts), do: opts
+
+  def call(conn, _) do
+    if conn.assigns[:current_user] != nil && conn.assigns[:current_user].role == :admin do
```
evansb: You can make an assumption that this Plug will be run after user is authenticated, otherwise it's programmer's error. `assert` is good for this.

indocomsoft:
All changes done.

lib/cadet_web/plug/check_admin.ex
```diff
@@ -0,0 +1,21 @@
+defmodule CadetWeb.Plug.CheckAdmin do
```
evansb: Sorry, can you rename this to `EnsureAdmin`? this is to be consistent with Guardian's `EnsureAuthenticated`

lib/cadet_web/plug/check_admin.ex
```diff
@@ -0,0 +1,21 @@
...
+  Checks whether :current_user is an admin.
+  If the user is not an admin, the HTTP 403 response will be sent.
+  """
+
+  import Plug.Conn
+
+  def init(opts), do: opts
+
+  def call(conn, _) do
+    if conn.assigns[:current_user].role == :admin do
```
evansb: Also I just realised this is wrong, the admin interface is meant to be opened by both Admin and Staff. So `EnsureStaff` will do

test/cadet_web/plug/check_admin_test.exs
```diff
@@ -0,0 +1,18 @@
...
+  use CadetWeb.ConnCase
+
+  alias CadetWeb.Plug.AssignCurrentUser
+  alias CadetWeb.Plug.CheckAdmin
+
+  test "init" do
+    CheckAdmin.init(%{})
+    # nothing to test
+  end
+
```
evansb: You also need to test for `:admin` and `:staff` right?

indocomsoft:
Instead I refactored them into `EnsureRoles` plug, I think this would be more versatile.

As for the test for when authenticated as admin and staff, the plug is supposed to 'do nothing', so I just `refute conn.status` (when there is no response yet, `conn.status` should be `nil` )

indocomsoft:
Responsibility handed over to frontend team @ningyuansg and @remo5000 in repo https://github.com/source-academy/cadet-frontend. Closing this.

PR #3, opened 2018-04-16
Add Credo to Travis and fix Credo offenses

indocomsoft:
Not sure whether we will still be using credo moving forward. But if we are, I added it to travis and fixed offenses in the current code base

### DIFF
--- a/lib/cadet/accounts/form/login.ex
+++ b/lib/cadet/accounts/form/login.ex
@@ -1,4 +1,7 @@
 defmodule Cadet.Accounts.Form.Login do
+  @moduledoc """
+  Validates login information provided through forms.
+  """
   use Ecto.Schema
 
   import Ecto.Changeset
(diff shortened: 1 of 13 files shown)


lib/cadet/application.ex
```diff
@@ -1,4 +1,7 @@
 defmodule Cadet.Application do
+  @moduledoc """
```
evansb: This module exist by Phoenix convention, it is expected for developers to understand what it contains, so `@moduledoc false` will do

lib/cadet/accounts/form/login.ex
```diff
@@ -1,4 +1,7 @@
 defmodule Cadet.Accounts.Form.Login do
+  @moduledoc """
+  Validates login information provided through forms.
```
evansb: This module (and the other form module) exist not only as validation but to implement the form entity itself, the validation is just something done for Ecto entities.
A better documentation would be to give a short description of what this entity represent (a login form comprises of e-mail and password)

indocomsoft:
Fixed according to your comments

PR #6, opened 2018-04-19
Add dialyzer static code analysis and fix dialyzer offenses

indocomsoft:


evansb:
Sorry, can you justify your decision of using Dialyzer?
1. It's very slow if you have large dependencies like Phoenix
2. It's neither sound nor complete
3. Having a good code coverage should suffice.

Furthermore it will cause more friction to the developers that have to learn/lookup the types.

indocomsoft:
Ah well, it was just a thought. I trust your judgment on this.

Although I've noticed some errors generated that might be legitimate errors like `changeset(entity, params \\ :empty)` with `:empty` instead of `%{}`, and some pattern matching in a `case` that is unreachable because of the pattern match in the parameter.

Maybe a better approach to run it once in a while just as a quick check and fix instead of imposing it on travis?

evansb:
I think that’s habit from Ecto 1, it should be %{} now. It used to be :empty to distinguish empty form

indocomsoft:
Alright, I'll fix stuffs that dialyxir complains about that are actually legit in this PR. Maybe keep `dialyxir` deps so that someone can run it once in a while? But I'll remove it from travis.

indocomsoft:
@evansb bump

PR #8, opened 2018-04-19
Implement polymorphism of question in Assessments.Mission entity

indocomsoft:
It is looking very complicated as of now. Maybe some ideas on how to simplify? Or should we revert back to the SA2 structure?


indocomsoft:
Will be redone according to Evan's intention.

PR #9, opened 2018-05-09
Fix 401 Unauthorized at base url for unauthenticated users

ningyuansg:
Previously, users who are not logged in (and therefore unauthenticated) would 
- receive a 401 unauthorized at `/`
- they would have to access `/session/new` to log in first
- after which they would be redirected to `/`. 

With these commits, users who are not logged in would
- be redirected to `/session/new` when visiting `/`, instead of receiving a 401 unauthorized

The behaviour for authenticated users are unchanged.

---

I'm not too certain if this is best practice (perhaps the logic could be implemented in the router instead of controller?)

### DIFF
--- a/lib/cadet/auth/error_handler.ex
+++ b/lib/cadet/auth/error_handler.ex
@@ -2,13 +2,18 @@ defmodule Cadet.Auth.ErrorHandler do
   @moduledoc """
   Handles authentication errors
   """
+  use CadetWeb, :controller
   import Plug.Conn
 
   def auth_error(conn, {type, _reason}, _opts) do
-    body = to_string(type)
+    if type == :unauthenticated do
+      conn |> redirect(to: session_path(conn, :new))
+    else
+      body = to_string(type)
 
-    conn
-    |> put_resp_content_type("text/html")
-    |> send_resp(401, body)
+      conn
+      |> put_resp_content_type("text/html")
+      |> send_resp(401, body)
+    end
   end
 end
--- a/lib/cadet/auth/error_handler.ex
+++ b/lib/cadet/auth/error_handler.ex
@@ -2,18 +2,13 @@ defmodule Cadet.Auth.ErrorHandler do
   @moduledoc """
   Handles authentication errors
   """
-  use CadetWeb, :controller
+  import Phoenix.Controller, only: [redirect: 2]
+  import CadetWeb.Router.Helpers, only: [session_path: 2]
   import Plug.Conn
 
-  def auth_error(conn, {type, _reason}, _opts) do
-    if type == :unauthenticated do
-      conn |> redirect(to: session_path(conn, :new))
-    else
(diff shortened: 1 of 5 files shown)


ningyuansg:
Looks like the drop in coverage is due to there not being a 401 unauthorized response in the tests. We can probably regain the coverage once the routes are more 'populated', so there are things to be unauthorized against?

Current routes,
```$ mix phx.routes
   page_path  GET   /                CadetWeb.PageController :index
session_path  GET   /session/new     CadetWeb.SessionController :new
session_path  POST  /session         CadetWeb.SessionController :create
session_path  GET   /session/logout  CadetWeb.SessionController :logout```

**Comment by evansb** [comment]
This should be implemented at middleware / Plug level instead of controller, otherwise every authenticated controller will have to implement the same logic.

Override the `auth_error` handler of Guardian, see 
https://hexdocs.pm/guardian/Guardian.Plug.EnsureAuthenticated.html

File: lib/cadet/auth/error_handler.ex [hunk 1]
```diff
@@ -2,13 +2,18 @@ defmodule Cadet.Auth.ErrorHandler do
   @moduledoc """
   Handles authentication errors
   """
+  use CadetWeb, :controller
   import Plug.Conn
 
   def auth_error(conn, {type, _reason}, _opts) do
-    body = to_string(type)
+    if type == :unauthenticated do
```
evansb: You don't need to handle the other case for now (which is `:already_authenticated`). It is used by `Guardian.Plug.EnsureNotAuthenticated`

evansb: So simply redirect, you can assume that `type` is always `:unauthenticated`

lib/cadet/auth/error_handler.ex
```diff
@@ -2,13 +2,18 @@ defmodule Cadet.Auth.ErrorHandler do
   @moduledoc """
   Handles authentication errors
   """
+  use CadetWeb, :controller
```
evansb: This is not a controller. If you want the `session_path` get it from `import CadetWeb.Router.Helpers, only: [session_path: 2]`

test/cadet_web/controllers/page_controller_test.exs
```diff
@@ -4,7 +4,7 @@ defmodule CadetWeb.PageControllerTest do
   describe "Unauthenticated User" do
     test "GET /", %{conn: conn} do
       conn = get(conn, "/")
-      assert html_response(conn, 401) =~ "unauthenticated"
+      assert redirected_to(conn) =~ "/session/new"
```
evansb: You should still check the error code.

lib/cadet/auth/error_handler.ex
evansb: Also the error code should still be 401

PR #31, opened 2018-05-14
Add PhoenixSwagger

indocomsoft:


indocomsoft:
@evansb I'm not sure how to test the swagger part since they are just "metadata" there.

evansb:
Create a test that simply use the function and does nothing, add a comment saying it's because swagger

indocomsoft:
Done.

Also a note to @sreycodes, `swagger_path` is a macro. The most relevant part from the source code is first line of this macro definition:
```elixir
defmacro swagger_path(action, [do: {:__block__, _, [first_expr | exprs]}]) do
      fun_name = "swagger_path_#{action}" |> String.to_atom
      body = Enum.reduce(exprs, first_expr, fn expr, acc ->
              quote do unquote(acc) |> unquote(expr) end
             end)

      quote do
        def unquote(fun_name)(route) do
```
So basically the macro creates another function with the name `swagger_path_create` when we do `swagger_path :create do` for example.

PR #32, opened 2018-05-19
Remove frontend

indocomsoft:
Fixes #27

Issue #26, opened 2018-05-12
Implement Changeset Test for Assessments Context


evansb:
Right now all of them have no tests. Follow https://github.com/source-academy/cadet/tree/master/test/cadet/course

indocomsoft:
Should be fixed by #34 when it is merged.

PR #33, opened 2018-05-19
Implement /auth in REST API

indocomsoft:
Based on #32, once that PR is merged, the no of files changed in this PR will be reduced significantly


indocomsoft:
@evansb cleaned up after #32 was merged. Currently the missing coverage is from the incomplete `Assessment` context (#26) and removal of authenticated pages (so the part of code that ensures authentication is not covered yet.)

lib/cadet_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,171 @@
...
+          conn
+          |> render("token.json", access_token: access_token, refresh_token: refresh_token)
+
+        {:error, _reason} ->
+          conn
+          |> send_resp(403, "Wrong email and/or password")
+      end
+    else
+      conn
+      |> send_resp(404, "Missing parameters")
```
evansb: Why 404 should be Bad Request

lib/cadet_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,171 @@
...
+    end
+  end
+
+  def create(conn, _params) do
+    conn
+    |> send_resp(404, "Missing parameters")
```
evansb: I think for one function call we shouldn't use pipe syntax, what do you think?

lib/cadet_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,171 @@
...
+        |> send_resp(401, "Invalid Token")
+    end
+  end
+
+  def refresh(conn, _params) do
+    conn
+    |> send_resp(404, "Missing parameter(s)")
+  end
+
+  def logout(conn, %{"refresh_token" => refresh_token}) do
```
evansb: Hmm why does logout need to pass the refresh token? After authenticating user you should simply delete the user entry in GuardianDB

lib/cadet_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,171 @@
...
+  end
+
+  def logout(conn, _params) do
+    conn
+    |> send_resp(404, "Missing parameter(s)")
+  end
+
+  swagger_path :create do
+    post("/auth")
+    summary("Obtain access and refresh tokens to authenticate user")
```
evansb: Since this is API documentation, this should be more descriptive so the frontend team knows what to do just by reading API docs. For instance, I think you should specify which HTTP header to put the token in, the token expiry time, etc

lib/cadet_web/router.ex
```diff
@@ -41,11 +40,12 @@ defmodule CadetWeb.Router do
...
         version: "1.0",
         title: "cadet"
-      }
+      },
+      basePath: "/v1"
     }
   end
 
-  scope "/api/swagger" do
+  scope "/v1/swagger" do
```
evansb: Typically we don't scope the swagger doc with API version because it's supposed to cover all versions.

lib/cadet_web/controllers/auth_controller.ex
indocomsoft: Sure. Maybe we can base our code styling on this guideline: https://github.com/christopheradams/elixir_style_guide ?

lib/cadet_web/controllers/auth_controller.ex
indocomsoft: Right, about that: we still haven't integrated guardiandb. Will integrate soon.

indocomsoft:
@evansb Re: logout needing to pass `refresh_token`
Because the hook is located on `Guardian.revoke(token)`, so I need to know the token to be invalidated (and since it's JWT this info is not stored anywhere by me). Access token is valid for 1 hour only, so it makes sense to simply invalidate the refresh token. However, if you like, I can let the user pass the access token and invalidate that as well.

Alternatively, I can add a field to user to store his/her `refresh_token` but I feel this solution is not as elegant since the client already stores the `refresh_token` anyway.

indocomsoft:
@evansb Fixed this according to your code review, should be ready for merging.

Issue #39, opened 2018-06-09
Grading Controller API Spec


indocomsoft:

indocomsoft:
Fixed by 72a8a1e00a232dfc0ce8b5d094114aafeba4b3fc

PR #41, opened 2018-06-09
Add API specs for Controllers

### LINKED ISSUE #40: Submission Controller API Spec

indocomsoft:
Fixes #40 #39 #43


remo5000:
Missions looks good so far, thanks for the quick PR!

For modularity of fetching data, will there be a GET request for individual missions implemented later on?

indocomsoft:
Yea, I'm thinking of giving less information in the missions listing, and then y'all can request more information about specific mission later on.

Which fields do you think is necessary for the missions index listing?

indocomsoft:
Yup, I got that, I'm just wondering which fields do you require for displaying information in the missions page? (e.g. You don't need question, but I'm not sure which other fields you need or don't need)

I think this is better so as to minimise the data sent. You can then get these additional informations when you request detailed data for a singular mission.

remo5000:
@indocomsoft We've come up with a mock of the data types that we would be displaying for assessments...the data is more or less similar to the model I saw in the swagger, barring a few instances of renaming. Also, we try to make a distinction between the different assessment types so as to conveniently fetch data according to the category later on and be more verbose

[https://gist.github.com/remo5000/cd6923c6fa004442bbc7c633718966fc](url)

lib/cadet_web/controllers/missions_controller.ex
```diff
@@ -27,10 +60,13 @@ defmodule CadetWeb.MissionsController do
       Mission:
         swagger_schema do
           properties do
+            order(:integer, "The order of showing the missions", required: true)
```
ningyuansg: Just to keep you in the know, we'll probably sort by due date or mission ID on the frontend (perhaps providing a toggle for the user). Still, no harm having this field here for back compatibility with mission data, but just to let you know.

remo5000:
I think Submission and GradingInfo look good. Just to clarify, for GradingInfo: 
- Is this a display-centric structure, i.e it combines data to return something for the front-end (that can be viewed when grading a submission)?
- If it is not, we could use it within `Grading`, to represent the grading of individual questions. I feel like there will be less repetition of data if we keep `questionId` instead of `weight`. (only caveat being the trouble of a foreign key).
- How would `weight` be calculated? In fact, we should decide on individual question marks or keeping the `maximumEXP`

As for Grading, I have this relationship in mind:
- **has a** Submission, Trainer, Student (both users), timestamp
- **has many** GradingInfos(?) to store how a trainer grades each question (i.e GradingInfo is not just for display). I feel like there is a better way to make this more ordered, but I'm not sure.

Any thoughts @ningyuansg @indocomsoft @sreycodes ?

indocomsoft:
- I'm trying my best to create a display-centric structure -- let me know if you require more or less.
- Right now @sreycodes and I are thinking of abolishing `weight` altogether, and instead grader would give xp for the whole submission (instead of individual question). This is assuming that there will be no graded MCQ. What do the front-end team think about this?

To be honest, the API structure that I made were rather display-centric so it doesn't quite correspond one-to-one to the backend models & relationships.

In case you're interested, here's the entity-relationship diagram of the `Assessments` context of backend. (`----<` means `has_many`, `- - - >` means `belongs_to`, and the text on the arrows is the field name represented by that relationship)
image

remo5000:
@indocomsoft  that looks good, I think the only modification when adding Grading would be shifting the `grader` field.

As for how to structure it, I think we should come up with structures that are easy to represent and modify. The API can then provide an ad-hoc structure using a combination of the simpler internal models. e.g the display while one is grading will take information from that particular `Submission`, `Grading` and `Question`.

I second removing `weight`, but do you think we could make do with individual `marks` for each `Question`? The total can then be dynamically counted by summing them up (when sending to an API request). So if, for example, a trainer decides to add a new question, just the question (and the marks it holds) would have to be added. What do you think?

indocomsoft:
@remo5000 not sure what you mean by "shifting the `grader` field". Could you elaborate more?

Yup, that's what I'm thinking. In such case, there's no need to create a separate table since the data needed have been contained in the models & contexts to be merged from the `assessments` branch. (#49 can be closed then).

As for removing `weight`, the backend team could certainly do that! It all depends on what workflow the frontend team is thinking of using. We can switch it to summing up total marks dynamically.

I will try to finish up my API spec for students submitting answers and graders giving comments and XP points by today so you and @ningyuansg can review the API tomorrow. If everything is okay, the frontend team can then write the necessary API hooks and the backend team can start implementing the controllers.

indocomsoft:
Alright, this is pretty much done. Let me know what you think @remo5000 @ningyuansg

remo5000:
@indocomsoft By moving the `grader` field, I mean decoupling a `Grading` (done by a trainer) from a `Submission` (done by a student). The changes would be just shifting the fields related to grading from the `Submission` model to the `Grading` model. This would make the schema more organised because we are [ensuring that each table holds specific data](https://en.wikipedia.org/wiki/Single_responsibility_principle). It also allows for the possibility of one-to-many relationship in the future (e.g for multiple people to grade a submission or to keep track of all the grading for a particular submission). I'll close the PR anyway, thanks for the speedy API specs 👍

indocomsoft:
@remo5000 Ok, I see what you mean now.

I guess for now we are still gonna have a `Submission` model with both `grader` and `student` field since the `assessments` branch is long overdue to be merged. Perhaps you could open an issue to refactor this in the future, once we are done implementing the basic functionalities.

lib/cadet_web/controllers/missions_controller.ex
```diff
@@ -0,0 +1,190 @@
...
+
+    response(200, "OK", Schema.ref(:Mission))
+    response(400, "Missing parameter(s) or invalid missionId")
+    response(401, "Unauthorised")
+  end
+
+  swagger_path :questions do
+    get("/missions/{missionId}/questions")
+
+    summary("Get questions contained inside a mission. Response is either `mcq` or `programming`")
```
remo5000: Does this mean that the response can contain (exclusive) either of the two? Right now the structure appears to be two arrays, one for `programming` and the other for `mcq`

lib/cadet_web/controllers/missions_controller.ex
```diff
@@ -0,0 +1,190 @@
...
+      missionId(:path, :integer, "mission id", required: true)
+    end
+
+    response(200, "OK", Schema.ref(:Mission))
+    response(400, "Missing parameter(s) or invalid missionId")
+    response(401, "Unauthorised")
+  end
+
+  swagger_path :questions do
+    get("/missions/{missionId}/questions")
```
ningyuansg: I don't think we need an API call for each question. Rather, the questions for a particular assessment can be provided with the API call for that assessment as an array. This helps us bind the right actions to the onClicks for next, prev, and submit (and when to render these buttons) of the workspace.

i.e. the `mission` model can supply us with questions as an array property.

lib/cadet_web/controllers/missions_controller.ex
```diff
@@ -0,0 +1,190 @@
...
+              :integer,
+              "The maximum amount of XP to be earned from this mission",
+              required: true
+            )
+
+            cover_picture(:string, "The URL to the cover picture", required: true)
+            mission_pdf(:string, "The URL to the mission pdf")
+          end
+        end,
+      Questions:
```
ningyuansg: Can we have a field in questions that is either 'MCQ' or 'Programming', or something to that effect? On our end, it's safer to check the string value of this property rather than something like `if (question.choices === undefined)`.

lib/cadet_web/controllers/missions_controller.ex
```diff
@@ -0,0 +1,190 @@
+defmodule CadetWeb.MissionsController do
```
ningyuansg: Do these missions include paths, contests, and sidequests? In that case, can we rename this as Assessments?

lib/cadet_web/controllers/missions_controller.ex
```diff
@@ -0,0 +1,190 @@
...
+                items(Schema.ref(:ProgrammingQuestion))
+              end
+            )
+          end
+        end,
+      MCQQuestion:
+        swagger_schema do
+          properties do
+            questionId(:integer, "question id", required: true)
+            content(:string, "the question content", required: true)
```
remo5000: Is this provided in Markdown? Or is it just the content/prompt of the question? @indocomsoft

lib/cadet_web/router.ex
```diff
@@ -28,6 +28,21 @@ defmodule CadetWeb.Router do
   # Authenticated Pages
   scope "/", CadetWeb do
     pipe_through([:api, :auth, :ensure_auth])
+
+    scope "/answer", Answer do
+      get("/programming/:id", ProgrammingAnswerController, :show)
+      post("/programming", ProgrammingAnswerController, :submit)
+      get("/mcq/:id", MCQAnswerController, :show)
+      post("/mcq", MCQAnswerController, :submit)
+    end
```
ningyuansg: As I understand it, question IDs are only unique within a mission, so this needs a mission ID somewhere.

Also, can we have one path for all questions, no matter the type (programming or mcq)? In that case, you could nest this under the current `/missions` path as the POST and GET at `/missions/{missionId}/{questionId}`.

lib/cadet_web/router.ex
```diff
@@ -28,6 +28,21 @@ defmodule CadetWeb.Router do
   # Authenticated Pages
```
ningyuansg: We need a path to get user information, such as username and (later on) progress on the story. Perhaps a `/user` path?

lib/cadet_web/controllers/answer/programming_answer_controller.ex
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def swagger_definitions do
+    %{
+      ProgrammingAnswer:
+        swagger_schema do
+          properties do
+            questionId(:integer, "The question id", required: true)
+
+            submitted(:boolean, "Whether this is the final answer", default: false)
```
tuesmiddt: Do we really need a submitted flag or can we just take the most recent submission as final?

lib/cadet_web/router.ex
tuesmiddt: I think we will probably need a full `users` controller. I can see the need to implement more features as we go along (notification preferences, display picture, whatnot)

lib/cadet_web/controllers/missions_controller.ex
tuesmiddt: It looks like one query to `/missions/{missionId}/questions` returns all the questions. However, you would then need to hit the `/answer/{type}/id` endpoint once per question to get the user's previous submission. Since we need to be able to refresh questions individually, I'm thinking maybe an alternative is to have `/answer/{type}?id={id1}?id={id2}` return a JSON of 
```
{
  id1: {
    type: "mcq",
    answer: 4
  },
  id2: {
    type: "programming",
    answer: "console.log('hello world!');"
  }, ...
}
```

ningyuansg: +1, though I'd prefer to not have to specify what type of question (MCQ or programming) in the query string.

tuesmiddt: You shouldn't need to specify the question type in the query. I guess it's not even necessary to include the type in the response. You would then need to check against the locally stored collection of questions to determine question type (and thus how to interpret the value).

Alternatively, for simplicity, we can represent MCQ answers as a string (specifically a character). So something like:
```
{
  id1: "d", // or "4"
  id2: "console.log('hello world!');"
}
```
@indocomsoft Thoughts?

lib/cadet_web/controllers/answer/programming_answer_controller.ex
indocomsoft: We need it cos students might wanna save them as "draft" -- once submitted, it's locked from being edited and the TA can start grading

lib/cadet_web/controllers/missions_controller.ex
indocomsoft: Yup, it's exclusive. Cos the swagger library we're using is still on OpenAPI 2.0 which doesn't allow `oneOf` 😞

lib/cadet_web/controllers/missions_controller.ex
indocomsoft: @tuesmiddt Yeah, sorry the `/answer/*` endpoints were my very early plans -- we should be able to unify the 2 types.

As for MCQ answer, I was thinking to maybe tag each MCQ choice associated with an MCQ question with a pseudo-id that will be used to identify them uniquely (combination of questionId + choice pseudo-id). Doesn't sound very nice to me -- let me know if any of you have any suggestion. @remo5000 @sreycodes @rrtheonlyone 

@ningyuansg would you then rather have the questions themselves embedded in the  `get("/missions/{missionId}")` endpoint? That way we can just merge the two.

lib/cadet_web/controllers/missions_controller.ex
indocomsoft: Sure. So the json will be either
```json
{
  "type": "mcq",
  "mcq": [
    { "questionId": 1,
      "content": "test",
      "choices": [{"content": "a", "hint": "asd"}]}]
    }
  ]
}
```
or
```json
{
  "type": "programming",
  "programming": [
    { "questionId": 2,
      "content": "kjnfb",
      "solution_template": "sjdfkb",
      "library": { "chapter": 1, "globals": [], "externals": [], "files": [] }
    }
  ]
}
```

Sounds good?

lib/cadet_web/controllers/missions_controller.ex
indocomsoft: It's just the prompt of the question, but we can store markdown, sure since it's just a string -- that depends on the frontend team decision

lib/cadet_web/router.ex
indocomsoft: question IDs should be unique across all but anyway this AnswerController needs a rework to unify the two kinds of questions anyway.

lib/cadet_web/controllers/missions_controller.ex
```diff
@@ -0,0 +1,190 @@
...
+
+    security([%{JWT: []}])
+
+    produces("application/json")
+
+    response(200, "OK", Schema.ref(:MissionsList))
+    response(401, "Unauthorised")
+  end
+
+  swagger_path :open do
```
indocomsoft: Remove this

lib/cadet_web/controllers/assessments_controller.ex
```diff
@@ -0,0 +1,163 @@
...
+          description("A list of all assessments")
+          type(:array)
+          items(Schema.ref(:AssessmentOverview))
+        end,
+      AssessmentOverview:
+        swagger_schema do
+          properties do
+            id(:integer, "The assessment id", required: true)
+            title(:string, "The title of the assessment", required: true)
+            type(:string, "Either mission/sidequest/path/contest", required: true)
```
indocomsoft: I think this should be `category`

indocomsoft: I think this should be `category`

tuesmiddt:
@ningyuansg @remo5000 The objects in the array returned by the `GET /assessments` endpoint no longer contain an `order` prop. I thought it's unnecessary since it's the array. Lmk if you do need it for other purposes.

lib/cadet_web/controllers/assessments_controller.ex
```diff
@@ -0,0 +1,163 @@
...
+              "The maximum amount of XP to be earned from this assessment",
+              required: true
+            )
+
+            cover_picture(:string, "The URL to the cover picture", required: true)
+          end
+        end,
+      Assessment:
+        swagger_schema do
+          properties do
```
indocomsoft: I think we should add `category` here just to standardise with the frontend data structure

lib/cadet_web/controllers/answer_controller.ex
```diff
@@ -0,0 +1,38 @@
+defmodule CadetWeb.AnswerController do
+  use CadetWeb, :controller
+
+  use PhoenixSwagger
+
+  swagger_path :submit do
+    post("/assessments/question/{questionId}/submit")
+
+    summary("Submit an answer to a question")
+
```
indocomsoft: I think you should add `description("For MCQ, answer contains choice_id. For programming question, this is a string containing the code.")`

lib/cadet_web/controllers/assessments_controller.ex
tuesmiddt: I didn't add this cos `AssessmentCategory` was in a different colour in the .ts schema. Need to get my eyes checked

indocomsoft: Lmao at least the commit message isn't `I'm blind`

indocomsoft:
LGTM

lib/cadet_web/controllers/assessments_controller.ex
```diff
@@ -0,0 +1,164 @@
...
+                type(:array)
+                items(Schema.ref(:MCQChoice))
+              end,
+              "MCQ choices if question type is mcq"
+            )
+
+            solution(:integer, "Solution to a mcq question if it belongs to path assessment")
+
+            answer(
+              :string_or_integer,
```
remo5000: @indocomsoft I was looking at the Grading specs and saw this. An issue with swagger perhaps? 
image

remo5000: Also, I think questionType and questionId should be changed to snake_case. I use camelCase for Typescript so that's probably where it got mixed in

PR #46, opened 2018-06-12
General cleanup

indocomsoft:


remo5000:
Looks good. May I suggest slightly expanding the scope of this PR? We can cleanup the README file in general: 
- Things like the frontend dependencies can be removed
- Also, will the URL for swagger be at `/v1/swagger` or just `/swagger`? 
- And the git version can be included as well in the PR

indocomsoft:
Sounds good. Will do the necessary changes

indocomsoft:
All done. @remo5000 @ningyuansg

PR #47, opened 2018-06-12
Initial structure for answer types

sreycodes:


test/cadet/assessments/answer_types/mcq_answer_test.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Cadet.Assessments.AnswerTypes.MCQAnswerTest do
+  use Cadet.ChangesetCase, async: true
+  
+  import Ecto.Changeset
+  alias Cadet.Assessments.AnswerTypes.MCQAnswer
+
+  test "valid changeset with correct choice" do
```
indocomsoft: Please use `valid_changesets` and `invalid_changesets` instead

lib/cadet/assessments/answer_types/mcq_answer.ex
```diff
@@ -0,0 +1,20 @@
...
+  It comprises of one of the MCQ choices.
+  """
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  alias Cadet.Assessments.QuestionTypes.MCQChoice
+
+  embedded_schema do
+    embeds_one(:answer_choice, MCQChoice)
```
indocomsoft: Actually now that I think about it, it might be better to give some kind of pseudo-id to our `MCQChoice` per `MCQQuestion`, so that this only needs to store the `MCQChoice`'s id.
So, e.g. for `MCQQuestion` with id 1, it has 4 choices, each with id's 1, 2, 3, and 4. Same thing for `MCQQuestion` with id 2, it has 4 choices, each with id's 1, 2, 3, and 4, and so on and so forth.
This will also make it easier for the API -- the client needs only submit the question id and choice id to submit an answer.
What do you think?

lib/cadet/assessments/answer_types/programming_answer.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Cadet.Assessments.AnswerTypes.ProgrammingAnswer do
+  @moduledoc """
+  The ProgrammingQuestion entity represents a Programming question.
+  """
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  embedded_schema do
+    field(:solution_code, :string)
```
indocomsoft: I think just naming it `:code` is better, since this is contained inside an **answer** entity already

test/cadet/assessments/answer_types/mcq_answer_test.exs
sreycodes: I couldn't get it to work for some weird reason.

lib/cadet/assessments/answer_types/mcq_answer.ex
sreycodes: Seems to be fine. The only downside I think is that we cannot trace question from an answer but such a case probably won't occur. I think we can go ahead with it.

test/cadet/assessments/answer_types/mcq_answer_test.exs
indocomsoft: What's the error message? Do ask on the group instead of coming of up with hackish solution like this.

Anyway since we're gonna rework this, it doesn't matter I guess

lib/cadet/assessments/answer_types/programming_answer.ex
indocomsoft: @sreycodes bump

test/cadet/assessments/answer_types/mcq_answer_test.exs
sreycodes: What is the problem with such a test though ?

lib/cadet/assessments/answer_types/programming_answer.ex
sreycodes: Done

lib/cadet/assessments/answer_types/mcq_answer.ex
sreycodes: Restructured but for standard 4 options. Should this be variable ?

.gitignore
```diff
@@ -20,7 +20,7 @@ erl_crash.dump
 
 # Uploaded file
 /uploads
-
+.DS_Store
 # Terraform Files
```
indocomsoft: Please add missing line break

lib/cadet/assessments/answer_types/mcq_answer.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Cadet.Assessments.AnswerTypes.MCQAnswer do
+  @moduledoc """
+  The Assessments.QuestionTypes.MCQQuestion entity represents an MCQ Answer.
+  It comprises of one of the MCQ choices.
+  """
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  alias Cadet.Assessments.QuestionTypes.MCQChoice
```
indocomsoft: Remove unused `alias`

lib/cadet/assessments/answer_types/mcq_answer.ex
```diff
@@ -0,0 +1,23 @@
...
+  embedded_schema do
+    field(:choice_id, :integer)  
+  end
+
+  @required_fields ~w(choice_id)a
+
+  def changeset(answer, params \\ %{}) do
+    answer
+    |> cast(params, @required_fields)
+    |> validate_number(:choice_id, greater_than: 0, less_than: 5)
```
indocomsoft: I don't think we should `validate_number` since it's possible in the future we have like 10 choices in MCQ. Let's not shoot ourselves in the foot by specifying hard limits.

lib/cadet/assessments/answer_types/mcq_answer.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  alias Cadet.Assessments.QuestionTypes.MCQChoice
+
+  embedded_schema do
+    field(:choice_id, :integer)  
+  end
+
+  @required_fields ~w(choice_id)a
+
+  def changeset(answer, params \\ %{}) do
```
indocomsoft: Please add `validate_required`

lib/cadet/assessments/question_types/mcq_choice.ex
```diff
@@ -12,14 +12,16 @@ defmodule Cadet.Assessments.QuestionTypes.MCQChoice do
...
 
-  @required_fields ~w(content is_correct)a
+  @required_fields ~w(content is_correct choice_id)a
   @optional_fields ~w(is_correct)a
 
   def changeset(question, params \\ %{}) do
     question
     |> cast(params, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
+    |> validate_number(:choice_id, greater_than: 0, less_than: 5)
```
indocomsoft: Same thing with `less_than`

test/cadet/assessments/question_types/mcq_question_test.exs
```diff
@@ -4,16 +4,16 @@ defmodule Cadet.Assessments.QuestionTypes.MCQQuestionTest do
   alias Cadet.Assessments.QuestionTypes.MCQQuestion
 
   valid_changesets MCQQuestion do
-    %{content: "asd", choices: [%{content: "asd", is_correct: true}]}
+    %{content: "asd", choices: [%{choice_id: 1, content: "asd", is_correct: true}]}
 
     %{
       raw_mcqquestion:
-        "{\"content\":\"asd\",\"choices\":[{\"is_correct\":true,\"content\":\"asd\"}]}"
+        "{\"content\":\"asd\",\"choices\":[{\"choice_id\": 1, \"is_correct\":true,\"content\":\"asd\"}]}"
```
indocomsoft: Use `Poison.encode!/1` on an elixir map instead -- looks neater

test/cadet/assessments/answer_types/mcq_answer_test.exs
indocomsoft: It's not DRY

PR #48, opened 2018-06-13
Implement /auth with IVLE token

ningyuansg:
### Features

- Remove email/password authentication
- Implement authentication with IVLE token
  1. Frontend sends an IVLE authentication token to /auth
  2. Backend receives the IVLE authentication token as `Cadet.Accounts.Form.Login`
  3. Backend uses IVLE authentication token to retrieve said token's corresponding NUSNET ID. This is also our outsourced 'authentication step'. IVLE api will return 200 if token is legitimate, otherwise 500.
  4. Every NUSNET ID is associated with a user (`Cadet.Accounts.User`). Backend then responds with signed `Tokens`.
  5. `/auth` will implicitly register the NUSNET ID into the database if not already registered.

### Todos

I'm used to developing incrementally, especially for languages and frameworks that are quite new to me; so I'd prefer to open separate issues/PRs for these.

- [ ] Check that the user reads CS1101S. The current implementation only checks that the user has an IVLE account.
- [ ] Retrieve the user's role in CS1101S and assign the role field accordingly. The register method currently hardcodes each new user as a `:student`. 
- [ ] Implement API call to retrieve name. This is an API call we will be using from frontend to backend immediately after login (there is a profile button we need to return upon login success that is labelled with the user's name).

### Dev notes

- The ivle api key is saved as an environment variable and retrieved as a module attribute. Meaning, you must set the appropriate `IVLE_KEY` value on compile time. This is reflected in the READMEs.
- You may come across an error such as `(Postgrex.Error) ERROR 42703 (undefined_column): column "token" of relation "authorizations" does not exist` on testing. This is because an early migration makes use of ecto_enum at `Cadet.Accounts.Provider`. Since this PR changes that Enum, and [apparently modifying enums is tricky (?)](https://github.com/gjaldon/ecto_enum#postgres), you will need to manually reset your test databases (`mix test` only migrates, not resets by default). Run the following:

```
$ export MIX_ENV=test
$ mix ecto.drop
$ mix ecto.create
$ mix ecto.migrate
$ mix run priv/repo/seeds.exs  # if you are at git root
$ unset MIX_ENV
```

If you do not unset `MIX_ENV`, you're gonna have a bad time (for that shell session at least).


ningyuansg:
I need some guidance on how to test functions that depend on external API calls (and functions that in turn depend on those). Referring to the module `Cadet.Accounts.Ivle` at `lib/cadet/accounts/ivle.ex'.

lib/cadet/accounts/ivle.ex
```diff
@@ -0,0 +1,77 @@
...
+  def fetch_name(token), do: api_fetch("UserName_Get", token)
+
+  defp api_fetch(path, token) do
+    # IVLE responds with 200 (:ok) if key and token are valid
+    # If key is invalid, IVLE responds with 500 (:internal_server_error)
+    # If token is invalid, IVLE returns an 'empty' string (i.e. "\"\"")
+    case HTTPoison.get(api_url(path, token)) do
+      {:ok, response} ->
+        if String.length(response.body) > 2 do
+          {:ok, String.replace(response.body, ~s("), "")}
```
remo5000: If there are other `"` characters in the response (apart from the start and end), I'm thinking it would be represented as `\"` as well. If this line replaces all the quotes, the `\` might remain. Is this the case? Not sure about the `~s(")` syntax here.

lib/cadet/accounts/accounts.ex
```diff
@@ -4,21 +4,21 @@ defmodule Cadet.Accounts do
...
-  authentication.
+  Register new User entity using Cadet.Accounts.Form.Registration
+
+  Returns {:ok, user} on success, otherwise {:error, changeset}
   """
   def register(attrs = %{}, role) do
     changeset = Registration.changeset(%Registration{}, attrs)
 
-    if changeset.valid?() do
+    if changeset.valid? do
```
ningyuansg: @indocomsoft The new style guide in #50 says that [zero-arity calls should end with parenthesis](https://github.com/lexmag/elixir-style-guide#zero-arity-parens), but I think predicate functions should be excluded, because their trailing `?` obviously indicates that they are predicate functions, not variables. Which do we go with? Your call.

indocomsoft: Well, in the first place, `changeset` is a map (https://hexdocs.pm/ecto/Ecto.Changeset.html#t:t/0), so `valid?` is just a key in the map, so there should definitely be no paranthesis here.

lib/cadet/accounts/ivle.ex
indocomsoft: `String.replace` looks a bit iffy -- how about `Poison.decode!(response.body) == ""` instead?

lib/cadet/accounts/ivle.ex
```diff
@@ -0,0 +1,77 @@
...
+      {:ok, "LEE NING YUAN"}
+
+  """
+  def fetch_name(token), do: api_fetch("UserName_Get", token)
+
+  defp api_fetch(path, token) do
+    # IVLE responds with 200 (:ok) if key and token are valid
+    # If key is invalid, IVLE responds with 500 (:internal_server_error)
+    # If token is invalid, IVLE returns an 'empty' string (i.e. "\"\"")
+    case HTTPoison.get(api_url(path, token)) do
```
indocomsoft: I think it's a good idea to use the `with` construct to prevent the pyramid of doom here as well. https://hexdocs.pm/elixir/Kernel.SpecialForms.html#with/1

lib/cadet_web/controllers/auth_controller.ex
```diff
@@ -6,35 +6,59 @@ defmodule CadetWeb.AuthController do
...
-      case Accounts.sign_in(login.email, login.password) do
-        {:ok, user} ->
-          {:ok, access_token, _} =
-            Guardian.encode_and_sign(user, %{}, token_type: "access", ttl: {1, :hour})
-
-          {:ok, refresh_token, _} =
-            Guardian.encode_and_sign(user, %{}, token_type: "refresh", ttl: {52, :weeks})
-
-          render(conn, "token.json", access_token: access_token, refresh_token: refresh_token)
+      case Ivle.fetch_nusnet_id(login.ivle_token) do
```
indocomsoft: `with` construct is better here to prevent pyramid of doom https://hexdocs.pm/elixir/Kernel.SpecialForms.html#with/1

lib/cadet_web/controllers/auth_controller.ex
```diff
@@ -84,8 +113,8 @@ defmodule CadetWeb.AuthController do
     end
 
     response(200, "OK", Schema.ref(:Tokens))
-    response(400, "Missing parameter(s)")
-    response(403, "Wrong login attributes")
+    response(400, "Missing or invalid parameter")
+    response(500, "Internal server error")
```
indocomsoft: Eh, this happens?

test/cadet/accounts/accounts_test.exs
```diff
@@ -0,0 +1,147 @@
...
+  end
+
+  test "create user" do
+    {:ok, user} =
+      Accounts.create_user(%{
+        name: "happy user",
+        role: :student
+      })
+
+    assert user.name == "happy user"
```
indocomsoft: Use pattern matching instead, e.g. `assert %{name: "happy user", role: :student} = user`. Similarly for the rest

test/cadet/accounts/accounts_test.exs
```diff
@@ -0,0 +1,147 @@
...
+    {:error, changeset} = Accounts.add_nusnet_id(user2, "teddy@happy.mail")
+    assert %{uid: ["has already been taken"]} = errors_on(changeset)
+  end
+
+  test "setting user nusnet_id without e-mail" do
+    user = insert(:user)
+    assert {:ok, []} = Accounts.set_nusnet_id(user, "E012345")
+  end
+
+  # TODO: A user may not have multiple NUSNET_IDs?
```
indocomsoft: I think it's a good idea to validate against this, but probably can be done in a separate PR. This PR is very big already haha

test/cadet_web/controllers/auth_controller_test.exs
```diff
@@ -1,89 +1,71 @@
...
+
   use CadetWeb.ConnCase
+  use ExVCR.Mock, adapter: ExVCR.Adapter.Hackney
 
   import Cadet.Factory
 
   alias CadetWeb.AuthController
   alias Cadet.Auth.Guardian
 
+  @token String.replace(inspect(System.get_env("TOKEN")), ~s("), "")
```
indocomsoft: Not sure what this does, why not a simple `System.get_env("TOKEN")`? Also what is this `TOKEN` environment variable?

README.md
```diff
@@ -21,10 +21,14 @@ Install Elixir dependencies
...
     mix deps.get
 
-Initialise Development Database
+Initialise development database
 
     mix ecto.setup
 
+Setup environment variables
+
+    export IVLE_KEY="1vl3keyy"
```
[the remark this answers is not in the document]
ningyuansg: No worries, not a real key. Moved this to the `.env` file with `5ab080e` and `3383e51`.

fixture/vcr_cassettes/accounts/sign_in#1.json
```diff
@@ -0,0 +1,29 @@
+[
```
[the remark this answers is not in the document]
ningyuansg: Done with `518307c`.

lib/cadet/accounts/accounts.ex
```diff
@@ -69,77 +68,59 @@ defmodule Cadet.Accounts do
...
-        {:ok, auth.user}
-    end
-  end
-
-  defp get_token(provider, uid) do
-    auth = Repo.get_by(Authorization, provider: provider, uid: uid)
+  def sign_in(nusnet_id, token) do
+    auth = Repo.one(Query.nusnet_id(nusnet_id))
 
     if auth == nil do
```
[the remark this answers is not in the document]
ningyuansg: Done with 2b892a6.

lib/cadet_web/controllers/auth_controller.ex
ningyuansg: Yep, IVLE responds with internal server error we send a wrong API key, and I think setting an invalid API key on our end is internal-server-error-esque.

test/cadet/accounts/accounts_test.exs
ningyuansg: Yep. Didn't know that this TODO would get pass credo though. Guess it ignores the test files.

lib/cadet/accounts/ivle.ex
ningyuansg: Refactor `String.replace` with `Poison.decode!` in 6120a44.

lib/cadet/accounts/accounts.ex
```diff
@@ -69,77 +68,59 @@ defmodule Cadet.Accounts do
...
-      nil
+      case Ivle.fetch_name(token) do
+        {:ok, name} ->
```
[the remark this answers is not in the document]
ningyuansg: Fixed excessive nesting using case with guards in 6120a44.

lib/cadet_web/controllers/auth_controller.ex
ningyuansg: Done with b2834fc.

test/cadet/accounts/accounts_test.exs
ningyuansg: Done with 735169f.

test/cadet_web/controllers/auth_controller_test.exs
ningyuansg: The token is the authentication token from ivle for the user. Added a brief description in the module docs of tests using `@token`, and improved the readability with 7075f93.

ningyuansg:
@indocomsoft Made the changes, rebased to master, let me know what you think. Thanks!

lib/cadet/accounts/ivle.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Cadet.Accounts.Ivle do
```
indocomsoft: (Sorry I missed this during the first review) I think the module name should be `Cadet.Accounts.IVLE` https://github.com/lexmag/elixir-style-guide#camelcase-modules

lib/cadet_web/controllers/auth_controller.ex
indocomsoft: Fair enough

.env
```diff
@@ -15,3 +15,6 @@ export CADET_WEBPACK_PORT=4001
 
 # Webpack entry filename
 export CADET_WEBPACK_ENTRY=app
```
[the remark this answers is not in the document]
ningyuansg: Done with 8d017f9.

.env
```diff
@@ -15,3 +15,6 @@ export CADET_WEBPACK_PORT=4001
```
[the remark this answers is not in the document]
ningyuansg: Done with 8d017f9, and README updated accordingly.

lib/cadet/accounts/accounts.ex
```diff
@@ -103,24 +103,27 @@ defmodule Cadet.Accounts do
...
+      # user is not registered in our database
+      with {:ok, name} <- Ivle.fetch_name(token),
+           {:ok, _} <- register(%{name: name, nusnet_id: nusnet_id}, :student) do
+        sign_in(nusnet_id, token)
+      else
+        {:error, :bad_request} ->
+          # Ivle.fetch_name/1 responds with :bad_request if token is invalid
+          {:error, :bad_request}
+
+        {:error, :internal_server_error} ->
```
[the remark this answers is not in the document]
ningyuansg: Fixed with 2fbf69e. Meant to be explicit, but I'll use comments.

lib/cadet/accounts/ivle.ex
```diff
@@ -54,19 +54,17 @@ defmodule Cadet.Accounts.Ivle do
...
+        {:ok, Poison.decode!(body)}
+
+      {:ok, %{status_code: 500}} ->
+        # IVLE responds with 500 if APIKey is invalid
         {:error, :internal_server_error}
+
+      {:ok, %{body: ~s(""), status_code: 200}} ->
+        # IVLE responsed 200 with body == ~s("") if token is invalid
+        {:error, :bad_request}
     end
```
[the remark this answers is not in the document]
ningyuansg: It's the desired behavior. The rationale is to prevent 'silent' failures. Frankly, I'm not sure what the best practice is with elixir phoenix, or api-building, so I'll go with your judgement on this one. Let me know

lib/cadet/accounts/ivle.ex
ningyuansg: Done with 429ca34. Do you prefer `IVLETest` or `IVLE_Test`?

test/fixture/vcr_cassettes/accounts/sign_in#1.json
```diff
@@ -16,10 +16,10 @@
         "Content-Type": "application/json",
```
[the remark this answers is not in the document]
ningyuansg: My bad, typo in the config file. Done with 04d2a1a.

lib/cadet/accounts/ivle.ex
```diff
@@ -9,7 +9,7 @@ defmodule Cadet.Accounts.Ivle do
   """
 
   @api_url "https://ivle.nus.edu.sg/api/Lapi.svc"
-  @api_key System.get_env("IVLE_KEY")
+  @api_key Dotenv.load!().values["IVLE_KEY"]
```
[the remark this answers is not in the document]
ningyuansg: Fixed with 706363b.

ningyuansg:
@indocomsoft Second round of fixes done, do take a look, thanks!

lib/cadet_web/controllers/auth_controller.ex
```diff
@@ -1,40 +1,59 @@
...
+      {:fetch, {:error, reason}} ->
+        # reason can be :bad_request or :internal_server_error
+        send_resp(conn, reason, "Unable to fetch NUSNET ID from IVLE.")
 
-        {:error, _reason} ->
-          send_resp(conn, :forbidden, "Wrong email and/or password")
-      end
-    else
-      send_resp(conn, :bad_request, "Missing parameters")
+      {:signin, {:error, reason}} ->
```
indocomsoft: This clause is missing test coverage hmm

ningyuansg: Done with eaa77fc.

lib/cadet/accounts/ivle.ex
```diff
@@ -0,0 +1,75 @@
...
+      {:ok, %{body: body, status_code: 200}} when body != ~s("") ->
+        {:ok, Poison.decode!(body)}
+
+      {:ok, %{status_code: 500}} ->
+        # IVLE responds with 500 if APIKey is invalid
+        {:error, :internal_server_error}
+
+      {:ok, %{body: ~s(""), status_code: 200}} ->
+        # IVLE responsed 200 with body == ~s("") if token is invalid
+        {:error, :bad_request}
```
indocomsoft: I think you should handle the case `{:error, _}`

test/cadet_web/controllers/auth_controller_test.exs
```diff
@@ -1,89 +1,90 @@
...
-        })
-
-      assert json_response(conn, 200)
+    test "invalid nusnet id", %{conn: conn} do
+      # an invalid nusnet id == ~s("") is typically caught by IVLE.fetch_nusnet_id
+      # the custom cassette skips this step so that we can test Accounts.sign_in
+      use_cassette "auth_controller/v1/auth#1", custom: true do
+        conn =
+          post(conn, "/v1/auth", %{
+            "login" => %{"ivle_token" => 'token'}
```
indocomsoft: Note that in elixir `'charlist'` and `"binary"` are 2 different ways of storing strings https://elixir-lang.org/getting-started/binaries-strings-and-char-lists.html

it should be `"token"` instead.

More to the point, it still doesn't cover: reading the report generated from running `mix coveralls.html test/cadet_web/controllers/auth_controller_test.exs:77`, the error is "caught" at `{:fetch, {:ok, nusnet_id}} <- {:fetch, IVLE.fetch_nusnet_id(login.ivle_token)}` since `api_fetch/2` returns `{:error, :bad_request}` on `%{body: ~s(""), status_code: 200}`.

One way to solve this might be to handle `{:error, _}` in `api_fetch/2`, then purposely fail the request `IVLE.fetch_name/1` if that makes sense.

lib/cadet/accounts/ivle.ex
ningyuansg: Gotcha, done with 0a9c589.

test/cadet_web/controllers/auth_controller_test.exs
ningyuansg: My bad, I should've checked the coveralls link. Fixed with 59a315f using custom cassettes. The fix didn't need the catch "don't care" clause, but I'll leave that change in.

lib/cadet/accounts/ivle.ex
```diff
@@ -0,0 +1,75 @@
...
+        {:ok, Poison.decode!(body)}
+
+      {:ok, %{status_code: 500}} ->
+        # IVLE responds with 500 if APIKey is invalid
+        {:error, :internal_server_error}
+
+      {:ok, _} ->
+        # IVLE responds 200 with body == ~s("") if token is invalid
+        {:error, :bad_request}
+    end
```
indocomsoft: Haha what I meant was adding another clause here `{:error, _} -> {:error, :internal_server_error}` or something like that, unless the intention is to let the server thread just crash on connection failure (which I think would also return HTTP 500)

ningyuansg: If they both return 500, I would rather let it crash. It'd be easier to catch unexpected bahaviour that way. Again, no experience with building APIs and web stuff here, so I'll default to your judgement. I've reverted it, but lmk what you'd prefer and I'll make the change.

indocomsoft:
Merged this for you -- maybe you could open issues in this PR's To-Do's

PR #72, opened 2018-06-25
Add login for guest account

ningyuansg:
## Features

- Programmatically obtain an authentication token for the IVLE guest user.


lib/cadet/public/updater.ex
```diff
@@ -0,0 +1,86 @@
...
+    response = HTTPoison.get!(@api_url_login)
+
+    viewstate =
+      response.body
+      |> Floki.find("input#__VIEWSTATE")
+      |> Floki.attribute("value")
+      |> List.first()
+
+    {"Set-Cookie", cookie} =
+      response.headers
```
indocomsoft: I think using `Enum.into/2` will look cleaner. (https://hexdocs.pm/elixir/1.2.6/Enum.html#into/2)
```elixir
cookie =
  response.headers
  |> Enum.into(%{})
  |> Map.get("Set-Cookie")
```

lib/cadet/public/updater.ex
```diff
@@ -0,0 +1,86 @@
...
+    with a 302. This time, an authentication token is embedded in the location,
+    which is extracted with `Regex`.
+  """
+  def get_token do
+    session = get_browser_session()
+    http_opts = [hackney: [cookie: session.cookie, follow_redirect: false]]
+    form = [userid: @username, password: @password, __VIEWSTATE: session.viewstate]
+
+    response = HTTPoison.post!(@api_url_login, {:form, form}, %{}, http_opts)
+    location = get_redirect_path(response)
```
indocomsoft: You can combine lines 31 and 32 by using pipe syntax
```elixir
location =
  @api_url_login
  |> HTTPoison.post!({:form, form}, %{}, http_opts)
  |> get_redirect_path()
```

lib/cadet/public/updater.ex
```diff
@@ -0,0 +1,86 @@
...
+        _ -> false
+      end)
+      |> List.first()
+
+    location
+  end
+
+  # Joins two URI components
+  defp join(uri1, uri2) do
+    URI.to_string(URI.merge(uri1, uri2))
```
indocomsoft: Please use pipe syntax
```elixir
uri1
|> URI.merge(uri2)
|> URI.to_string()
```

lib/cadet/public/updater.ex
```diff
@@ -0,0 +1,86 @@
...
+        {"Location", _} -> true
+        _ -> false
+      end)
+      |> List.first()
+
+    location
+  end
+
+  # Joins two URI components
+  defp join(uri1, uri2) do
```
indocomsoft: Is this function used once only? If so I don't really see the point in extracting this out

lib/cadet/public/updater.ex
```diff
@@ -0,0 +1,84 @@
...
+      |> HTTPoison.post!({:form, form}, %{}, http_opts)
+      |> get_redirect_path()
+
+    response =
+      @api_url
+      |> URI.merge(location)
+      |> URI.to_string()
+      |> HTTPoison.get!(%{}, http_opts)
+
+    token =
```
indocomsoft: Instead of doing
```elixir
token =
  response
  |> some_function
  |> more_function

token
```

You can just do
```elixir
response
|> some_function
|> more_function
```

This is because elixir will return the value of the last expression in a function. Similar design pattern as in Ruby

indocomsoft: Yeah, this is fine. The name of the function is `get_token` anyway, it should be obvious enough.

ningyuansg: Done with 110dae3.

PR #73, opened 2018-06-25
Implement role fetching

ningyuansg:
## Features

- Add a function in the `IVLE` module to check the role of the user
  - While waiting for confirmation as to what role each 'Permission' key in the response JSON corresponds to, I'm assuming `O` means owner, `S` is student, and everything in-between is staff.
- Refactor the private `api_url` and `api_fetch` functions
  - Had to make them more flexible since some api methods received token as parameter `Token`, while others used `AuthToken`.


indocomsoft:
@ningyuansg Sorry, I was on mobile, merged the wrong PR. Could you help re-create the PR?

ningyuansg:
Done with #76

PR #74, opened 2018-06-27
Implement Updater as a GenServer

ningyuansg:
## Features

- Cadet.Public.Updater is started by passing the command line flag `--updater`
  - Using mix, run `elixir --erl "--updater" -S mix phx.server`. (or `iex`)
  - Using the compiled binary, run `bin/cadet start --updater`
- Cadet.Public.Updater is a GenServer that runs as a worker in the Cadet.Application supervision tree
- Cadet.Public.Updater will periodically do work defined in `handle_info/2(:work, ...)`. 
- Cadet.Public.Updater schedules itself recursively by a private function `schedule_work/0`, which calls `handle_info/2(:work, ...)`, which calls `schedule_work/0`, and so on
- The interval at which this happens is defined in the config file (config/config.exs)


indocomsoft:
I think you should rebase this on `master` first to reduce the number of changes @ningyuansg

ningyuansg:
@indocomsoft Rebased, but you'd probably want to look at #76 first, since this PR uses _one_ commit from that branch (90497ee).

test/cadet/public/updater_test.exs
```diff
@@ -31,4 +34,71 @@ defmodule Cadet.Public.UpdaterTest do
       assert String.length(token) == 480
     end
   end
+
+  test "Get course id" do
+    use_cassette "updater/get_course_id#1", custom: true do
+      course_id = Updater.get_course_id("T0K3N...")
+      assert String.length(course_id) == 36
+      assert course_id != "00000000-0000-0000-0000-000000000000"
```
indocomsoft: Should be `refute course_id == "00000000-0000-0000-0000-000000000000"`

test/cadet/public/updater_test.exs
```diff
@@ -31,4 +34,71 @@ defmodule Cadet.Public.UpdaterTest do
...
+      course_id = Updater.get_course_id(token)
+      assert {:ok, announcements} = Updater.get_announcements(token, course_id)
+      assert is_list(announcements)
+    end
+  end
+
+  describe "Start GenServer" do
+    test "Using GenServer.start_link/3" do
+      use_cassette "updater/start_link#1", custom: true do
+        assert {:ok, _} = GenServer.start_link(Updater, %{}, name: TestUpdater)
```
indocomsoft: If `Updater.init/1` ignores its argument, then change `%{}` to `nil`

test/cadet/public/updater_test.exs
```diff
@@ -31,4 +34,71 @@ defmodule Cadet.Public.UpdaterTest do
...
+        assert :ok = GenServer.stop(TestUpdater)
+      end
+    end
+
+    test "Using Updater.start_link/1" do
+      use_cassette "updater/start_link#2", custom: true do
+        assert {:ok, pid} = Updater.start_link()
+        assert Process.alive?(pid) == true
```
indocomsoft: `== true` is superfluous, can be removed

test/cadet/public/updater_test.exs
```diff
@@ -31,4 +34,71 @@ defmodule Cadet.Public.UpdaterTest do
...
+        assert GenServer.stop(pid) == :ok
+        assert Process.alive?(pid) == false
```
indocomsoft: Instead do `refute Process.alive?(pid)`

test/cadet/public/updater_test.exs
```diff
@@ -31,4 +34,71 @@ defmodule Cadet.Public.UpdaterTest do
...
+      end
+    end
+  end
+
+  test "GenServer init/1 callback" do
+    use_cassette "updater/init#1", custom: true do
+      assert {:ok, %{token: token, course_id: course_id}} = Updater.init(%{})
```
indocomsoft: Change `%{}` to `nil` if it is ignored

test/cadet/public/updater_test.exs
```diff
@@ -31,4 +34,71 @@ defmodule Cadet.Public.UpdaterTest do
...
+      assert String.length(token) == 480
+      assert String.length(course_id) == 36
+      assert course_id != "00000000-0000-0000-0000-000000000000"
```
indocomsoft: use `refute`

test/cadet/public/updater_test.exs
```diff
@@ -31,4 +34,71 @@ defmodule Cadet.Public.UpdaterTest do
...
+    end
+  end
+
+  describe "Send :work to GenServer" do
+    test "Valid token" do
+      use_cassette "updater/handle_info#1", custom: true do
+        token = Updater.get_token()
+        course_id = Updater.get_course_id(token)
+        api_params = %{token: token, course_id: course_id}
```
indocomsoft: Once `get_api_params/0` is refactored, use that instead here.

test/cadet/public/updater_test.exs
```diff
@@ -31,4 +34,71 @@ defmodule Cadet.Public.UpdaterTest do
...
+        assert {:noreply, new_api_params} = Updater.handle_info(:work, api_params)
+        assert new_api_params == api_params
+      end
+    end
+
+    test "Invalid token" do
+      use_cassette "updater/handle_info#2", custom: true do
+        token = Updater.get_token()
+        course_id = Updater.get_course_id(token)
+        api_params = %{token: "bad_token", course_id: course_id}
```
indocomsoft: `get_api_params/0`

lib/cadet/accounts/ivle.ex
```diff
@@ -63,13 +160,23 @@ defmodule Cadet.Accounts.IVLE do
...
+  # The authentication token parameter must be provided explicitly rather than
+  # provided implicitly by this function as some API calls use ...&Token={token},
+  # while others use ...&AuthToken={token}
+  defp api_url(method, queries) do
+    queries = [APIKey: @api_key] ++ queries
+
+    url =
+      @api_url
+      |> URI.merge(method)
+      |> (&"#{&1}?#{URI.encode_query(queries)}").()
```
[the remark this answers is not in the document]
ningyuansg: Fixed with rebase to #76.

config/config.exs
```diff
@@ -6,7 +6,9 @@
 use Mix.Config
 
 # General application configuration
-config :cadet, ecto_repos: [Cadet.Repo]
+config :cadet,
+  ecto_repos: [Cadet.Repo],
+  updater: [interval: 5 * 60 * 1000]
```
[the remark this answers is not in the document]
ningyuansg: Can't easily use Timex in the config, since config is evaluated before dependencies are compiled. I don't think it's worth the trouble to implement a work around just to be able to use this pipe. Added a comment indicating that the value for `interval` is in milliseconds in fd9de8e.

lib/cadet/public/updater.ex
```diff
@@ -6,12 +6,111 @@ defmodule Cadet.Public.Updater do
...
+  @doc """
+  Callback for the GenServer. This function calls `schedule_work`, which
+  initiates a recursive call every `@interval` milliseconds. This acts as a
+  regularly scheduled task (e.g. cronjob).
+
+  `start_link/0` -> `init/1` -> `schedule_work/0` -> `handle_info/2` ->
+    `schedule_work/0` -> `handle_info/2` -> ...
+  """
+  def init(_) do
+    token = get_token()
```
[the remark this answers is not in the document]
ningyuansg: Refactored with 9131740.

lib/cadet/accounts/ivle.ex
```diff
@@ -51,10 +50,108 @@ defmodule Cadet.Accounts.IVLE do
...
+  `"00000000-0000-0000-0000-000000000000"`, and that there is only one active
+  module with the course code `"CS1101S"`. (So far, these assumptions have been
+  true).
+  """
+  def fetch_role(token) do
+    {:ok, modules} = api_call("Modules", AuthToken: token, CourseID: "CS1101S")
+
+    cs1101s =
+      modules["Results"]
+      |> Enum.filter(fn module ->
```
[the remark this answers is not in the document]
ningyuansg: Done with e57057c.

lib/cadet/public/updater.ex
```diff
@@ -6,12 +6,111 @@ defmodule Cadet.Public.Updater do
...
+
+    - token: String, the IVLE authentication token
+
+  """
+  def get_course_id(token) do
+    {:ok, modules} = IVLE.api_call("Modules", AuthToken: token, CourseID: "CS1101S")
+
+    cs1101s =
+      modules["Results"]
+      |> Enum.filter(fn module ->
```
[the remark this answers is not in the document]
ningyuansg: Done with e57057c.

test/cadet/public/updater_test.exs
ningyuansg: Done with a2ae3d9.

test/cadet/public/updater_test.exs
ningyuansg: Done with a2ae3d9.

test/cadet/public/updater_test.exs
ningyuansg: Done with a2ae3d9.

test/cadet/public/updater_test.exs
ningyuansg: Done with a2ae3d9.

test/cadet/public/updater_test.exs
ningyuansg: Done with a2ae3d9.

test/cadet/public/updater_test.exs
ningyuansg: Done with a2ae3d9.

test/cadet/public/updater_test.exs
ningyuansg: Done with a2ae3d9.

test/cadet/public/updater_test.exs
ningyuansg: Done with a2ae3d9.

ningyuansg:
Adding tests for new `get_api_params/0`. Please wait on merge.

ningyuansg:
Test coverage fell, seems `updater/handle_info#2` _(Send :work to GenServer, Invalid token)_ is not working as expected. Please wait on merge.

ningyuansg:
@indocomsoft We need to hold off on this merge until I implement some sort of flag that decides if the GenServer should be run. The issue is there are at least two ec2 instances running this application simultaneously because of the AWS auto-scaling group. As a result, there would be at least twice the expected number of calls to IVLE due to this GenServer. 

This is also a problem when we have to implement automated grading later on. Only one 'main' instance of this application should be the one checking the deadline and sending out requests to the AWS lambda auto grader. Otherwise, we end up duplication calls.

ningyuansg:
@indocomsoft Rebased and adding command line options to start the updater.

lib/cadet/accounts/ivle.ex
```diff
@@ -107,7 +109,60 @@ defmodule Cadet.Accounts.IVLE do
...
+    - queries: [Keyword], key-value pair of parameters to send
+
+  This method is valid for methods that return a 200 with an empty string body
+  on invalid tokens. For methods that return with string "Invalid login!" in a
+  JSON nested in the body, refer to the previous method api_call/2 with guard
+  clause.
+  """
+  def api_call(method, queries) do
     case HTTPoison.get(api_url(method, queries)) do
       {:ok, %{body: body, status_code: 200}} when body != ~s("") ->
```
indocomsoft: Instead of doing
```elixir
{:ok, %{body: body, status_code: 200}} when body != ~s("") ->
{:ok, %{body: ~s(""), status_code: 200}} ->
```
Maybe you could consider doing 
```elixir
{:ok, %{body: ~s(""), status_code: 200}} ->
{:ok, %{body: body, status_code: 200}} ->
```

Since `case` pattern matching is evaluated sequentially

indocomsoft:
Might wanna add notice on README to warn people that changing `.env` requires a re-compilation with  `mix clean` because module attribute is only evaluated during compile time

lib/cadet/application.ex
```diff
@@ -19,6 +19,18 @@ defmodule Cadet.Application do
       worker(Guardian.DB.Token.SweeperServer, [])
     ]
 
+    # To supply command line args to phx.server, you must use the elixir/iex bin
+    #     $ elixir --erl "--updater" -S mix phx.server
+    #     $ iex --erl "--updater" -S mix phx.server
```
indocomsoft: Should be `cadet.server` me thinks

lib/cadet/accounts/ivle.ex
```diff
@@ -82,15 +85,14 @@ defmodule Cadet.Accounts.IVLE do
 
   """
   def fetch_role(token) do
-    {:ok, modules} = api_fetch("Modules", AuthToken: token, CourseID: "CS1101S")
+    {:ok, modules} = api_call("Modules", AuthToken: token, CourseID: "CS1101S")
 
     cs1101s =
```
indocomsoft: Not sure if the 
`cs1101s = modules["Results"] |> Enum.find` part should be refactored to a separate method. Currently there are only 2 occurrences of them, so let's keep them separate for now following the rule of 3

lib/cadet/application.ex
ningyuansg: It's aliased to `phx.server` at [`mix.exs`](https://github.com/source-academy/cadet/blob/896abe5e204b903745fb29f24fbd180140dc9e8c/mix.exs#L86)

lib/cadet/accounts/ivle.ex
ningyuansg: I prefer the current match order, since it follows an 'if ok, else error' structure, rather than an 'if error, elif ok, else error'.

ningyuansg:
@indocomsoft There's already instructions about recompilation in the current README.

> If you've compiled the application before setting a valid value, you must force a recompilation with `mix clean && mix`.

PR #77, opened 2018-06-29
Add package distillery

ningyuansg:
### Features

- Add package distillery
  - Also used in the previous iteration of source academy
  - Allows bundling of erlang runtime, simplifies deployment to production servers
- Move git_hooks config from global config to dev config file
  - Allows us to avoid bundling this package together with the production build
- Set key server to true for CadetWeb.Endpoint
  - This is required for the bundled production build to work

### Todos

Quite a lot. Will self-host a temporary simple 'staging' server on my DO instance while I figure out

- [ ] Add a postgres RDS to the staging vpc via terraform
- [ ] Figure out how to connect the ec2 instances to RDS
- [ ] Use packer to generate appropriate AMIs
- [ ] Get terraform to make use of those packer-generated AMIs


rel/config.exs
```diff
@@ -0,0 +1,53 @@
...
+  # dev mode.
+  set dev_mode: true
+  set include_erts: false
+  set cookie: :"pDlr~SGoSLjoW12Z}hefp7vH!nN[0?Pj=,W6:>fkB/Hvjd=6)uE<*JiZ~E5(DpK7"
+end
+
+environment :prod do
+  set include_erts: true
+  set include_src: false
+  set cookie: :"l^m/AL2P4|/~aF)&j|j8<PT1bRuAA*kYEA`8L0XQuh=5NVkuu%.0^PVPXBVgtKFL"
```
indocomsoft: Security concern: Erlang magic cookie shouldn't be hardcoded. Probably take from `dotenv` instead

This is probably a good reference: https://hackernoon.com/mastering-elixir-releases-with-distillery-a-pretty-complete-guide-497546f298bc

indocomsoft:
LGTM

PR #81, opened 2018-07-02
Add postgres RDS via terraform

ningyuansg:
### Features

- The RDS is in a private subnet, and so for now it should not matter that the username and password are public
  - Nonetheless, it _should_ be hidden for production. But we can deal with that later
- Change the ssh key because only one key pair can be associated as the 'start-up' keypair for an AWS ec2 instance and I don't believe I have the private key for the old one.


terraform/cadet/rds.tf
```diff
@@ -0,0 +1,28 @@
...
+  engine                 = "postgres"
+  username               = "postgres"
+  password               = "postgres"
+  port                   = 5432
+  publicly_accessible    = false
+
+  tags {
+    Name        = "${title(var.env)} Cadet DB"
+    Environment = "${var.env}"
+  }
```
evansb: You might want to change the password later using console. In this case it might be good to have `lifecycle { ignore_changes = ["password"] }` block.

terraform/cadet/vpc.tf
```diff
@@ -7,10 +7,32 @@ resource "aws_vpc" "cadet" {
   }
 }
 
+resource "aws_subnet" "private_a" {
```
evansb: I think there is a requirement for the subnets to have different AZs, otherwise high availability is impossible

terraform/cadet/vpc.tf
```diff
@@ -7,10 +7,32 @@ resource "aws_vpc" "cadet" {
...
+  }
+}
+
+resource "aws_subnet" "private_b" {
+  vpc_id                  = "${aws_vpc.cadet.id}"
+  cidr_block              = "10.0.1.0/24"
+  map_public_ip_on_launch = false
+
+  tags {
+    Name        = "${title(var.env)} Cadet Private Subnet A"
```
evansb: So this should be subnet B and explicitly use `ap-southeast-1b`

terraform/cadet/rds.tf
```diff
@@ -0,0 +1,28 @@
...
+}
+
+resource "aws_db_instance" "db" {
+  name                   = "${title(var.env)}CadetDB"
+  instance_class         = "db.t2.micro"
```
evansb: You might want this to be configurable via inputs

evansb:
Note that Amazon Block Storage IOPS is proportional to its storage limit, Read: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#Concepts.Storage.GeneralSSD, so sometimes you need to over-provision the storage capacity just to get the desired IOPS

ningyuansg:
@evansb Thanks! Changes made with d2cb217:

- Use input variable for RDS instance class
  - As a result, renamed instance_type -> ec2_instance_type to differentiate between ec2 and rds
- Use input variable for RDS allocated storage
- Add lifecycle block for RDS
- Specify explicitly the AZ for the DB subnet group
- Fix typo with private subnet B tag name

indocomsoft:
As requested by @ningyuansg, merging this first while issues brought up in this PR will be dealt with in a separate PR, tracked by #95

PR #86, opened 2018-07-04
Update answers model and add migration

tuesmiddt:
- Add answers table
- Remove question type from answer model


indocomsoft:
I wonder if there's a way to verify structure using AnswerTypes somehow in the changeset

tuesmiddt:
I don't think we can use `cast_embed/2` since our schema defines `answer` as a field and does not embed an answer of a specific type. We can't use `embeds_one/3` in the schema since Ecto does not allow polymorphic embeds. 

I'll write a custom validator in the answers controller branch. In the meantime you just have to ensure that it's built correctly in your tests.

PR #106, opened 2018-07-09
Mock ivle_token to allow easy sign in in dev environment

### LINKED ISSUE #104: Mock ivle_token to allow easy sign in in dev environment

indocomsoft:
Fixes #104

### DIFF
--- a/lib/mix/tasks/token.ex
+++ b/lib/mix/tasks/token.ex
@@ -45,7 +45,7 @@ defmodule Mix.Tasks.Cadet.Token do
       IO.puts("#{bright()}Test user id:#{reset()} #{cyan()}#{user.id}#{reset()}")
       IO.puts("#{bright()}Test user:#{reset()}")
       IO.puts("#{cyan()}#{inspect(user, pretty: true)}#{reset()}")
-      IO.puts("#{bright}JWT:#{reset}")
+      IO.puts("#{bright()}JWT:#{reset()}")
       IO.puts("Bearer #{access_token}")
     else
       IO.puts("Invalid arguments provided.")
--- a/lib/mix/tasks/token.ex
+++ b/lib/mix/tasks/token.ex
@@ -1,0 +1,67 @@
+defmodule Mix.Tasks.Cadet.Token do
+  @moduledoc """
+  Helper to generate access_token to ease development.
+
+  Usage: `mix task cadet.token <role>`
+
+  where <role> in #{inspect(Enum.filter(Cadet.Accounts.Role.__valid_values__(), &is_binary/1))}
+
+  For example: `mix task cadet.token student`
+  """
+
+  @shortdoc "Generates access_token JWT for a given user role"
+
+  use Mix.Task
+
+  import Mix.Ecto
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/token.ex
```diff
@@ -0,0 +1,64 @@
...
+  @env_allow_mock ~w(dev)a
+
+  def run(args) do
+    # Suppress Ecto SQL logs
+    Logger.configure(level: :error)
+
+    # Required for Ecto to work properly, from Mix.Ecto
+    ensure_started(Repo, [])
+
+    roles = Enum.filter(Role.__valid_values__(), &is_binary/1)
```
tuesmiddt: Can replace with `roles = Role.__enum_map__()`

lib/mix/tasks/token.ex
```diff
@@ -0,0 +1,64 @@
...
+    role = List.first(args)
+
+    if Enum.count(args) == 1 and role in roles do
+      {:ok, access_token, _} =
+        Guardian.encode_and_sign(test_user(role), %{}, token_type: "access", ttl: {4, :weeks})
+
+      IO.puts(access_token)
```
tuesmiddt: Since this is a convenience task, I'm tempted to print `Bearer <token>` instead but your call.

lib/mix/tasks/token.ex
indocomsoft: That would be atoms tho.

I could also do `Enum.map(Role.__enum_map__(), &Atom.to_string/1)`

lib/mix/tasks/token.ex
indocomsoft: Hmm the problem with that is, `Compiling file` message is actually printed in stdout, so you can't just do a simple pipe to `xclip` or `pbcopy`

```bash
650 (mock-dev-user) $ mix cadet.token student 2>/dev/null
Compiling 1 file (.ex)
eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJjYWRldCIsImV4cCI6MTUzMzU3MjM1MywiaWF0IjoxNTMxMTUzMTUzLCJpc3MiOiJjYWRldCIsImp0aSI6IjY4NzQ4NDMxLTUzNzAtNDg3Mi1hMzBkLTMwMmZmMjcwZDQ0OSIsIm5iZiI6MTUzMTE1MzE1Miwic3ViIjoiMSIsInR5cCI6ImFjY2VzcyJ9.TFIVEdG64goERJ_dZUGcS7xxCehc3JxFj3-_ZbjkQXjYyO6AM5diP05QUwUwmtfvRpjTGLCZ5Dt4a2XCKTo9mw
```

So I think I'll keep at printing `<token>`

indocomsoft: Actually you can pipe it to `tail -n1` so I guess I'll print `Bearer <token>` then

lib/mix/tasks/token.ex
tuesmiddt: derped. mb

PR #107, opened 2018-07-10
Rename Cadet.Public.Updater -> Cadet.Updater.Public

ningyuansg:
In anticipation for other scheduled tasks (more 'updaters'), make Public a subset of these scheduled tasks, rather than the other way round.


lib/cadet/updater/updater.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Cadet.Public.Updater do
+defmodule Cadet.Updater.Public do
```
indocomsoft: I think you meant the file name to be `lib/cadet/updater/public.ex`?

ningyuansg: oh no

indocomsoft:
LGTM

PR #109, opened 2018-07-11
Add schedulers pulling from cs1101s git repo

ningyuansg:
I'm starting to think that the `.env` file and the `config/prod.secret.exs` files aren't so different. Kind of looks like the `.env` is a relic of a react frontend. Not particularly important, but could we open an issue moving configuration from `.env` to `config/*.exs`? Think it might be best practice.


lib/cadet/updater/CS1101S.ex
```diff
@@ -0,0 +1,69 @@
...
+    git("pull")
+  end
+
+  defp git(cmd, args \\ [])
+
+  defp git("clone", args) do
+    {out, exit} =
+      System.cmd(
+        "git",
+        ["clone"] ++ args,
```
indocomsoft: this one does not have the `-C @local_name` argument?

test/support/data_case.ex
```diff
@@ -51,4 +53,28 @@ defmodule Cadet.DataCase do
       end)
     end)
   end
+
+  @doc """
+  Run a git command with `loc` as the git repository root.
+  """
+  def git_from(loc, cmd, opts \\ []) do
```
indocomsoft: If the following functions are used only by the deployment test, I suggest moving this either to the test file directly as a `defp` or a separate file `updater_helper.exs`

test/cadet/updater/CS1101S_test.exs
```diff
@@ -0,0 +1,44 @@
...
+  use ExUnit.Case, async: false
+
+  alias Cadet.Updater.CS1101S
+
+  setup_all do
+    {_, 0} = System.cmd("mkdir", [@remote_repo])
+    {_, 0} = git_from(@remote_repo, "init", ["--bare"])
+    # git pull fails on an empty --bare repo, so need to push something there first
+    git_add_file("dummy_setup_all", @remote_repo)
+    on_exit(fn -> clean_dirs([@local_repo, @remote_repo]) end)
```
indocomsoft: Prefer `& clean_dirs([@local_repo, @remote_repo])`

test/support/data_case.ex
```diff
@@ -51,4 +53,28 @@ defmodule Cadet.DataCase do
...
+    {_, 0} = git_from(@temp_repo, "add", [filename])
+    {_, 0} = git_from(@temp_repo, "commit", ["-m", "dummy-commit"])
+    {_, 0} = git_from(@temp_repo, "push")
+    {_, 0} = clean_dirs([@temp_repo])
+  end
+
+  @doc """
+  Remove directories.
+  """
+  def clean_dirs(dirs), do: System.cmd("rm", ["-rf"] ++ dirs)
```
indocomsoft: Prefer using https://hexdocs.pm/elixir/File.html#rm_rf/1

test/support/data_case.ex
```diff
@@ -51,4 +53,28 @@ defmodule Cadet.DataCase do
...
+  def git_from(loc, cmd, opts \\ []) do
+    System.cmd("git", ["-C", loc, cmd] ++ opts, stderr_to_stdout: true)
+  end
+
+  @doc """
+  Push an empty file with `filename` into the given remote repository
+  """
+  def git_add_file(filename, remote_repo) do
+    {_, 0} = git_from(".", "clone", [remote_repo, @temp_repo])
+    {_, 0} = System.cmd("touch", [Path.join(@temp_repo, filename)])
```
indocomsoft: Prefer using https://hexdocs.pm/elixir/File.html#touch/2

test/cadet/updater/CS1101S_test.exs
```diff
@@ -0,0 +1,44 @@
...
+  @remote_repo "test/remote_repo"
+  @local_repo "test/local_repo"
+
+  use Cadet.DataCase
+  use ExUnit.Case, async: false
+
+  alias Cadet.Updater.CS1101S
+
+  setup_all do
+    {_, 0} = System.cmd("mkdir", [@remote_repo])
```
indocomsoft: Prefer https://hexdocs.pm/elixir/File.html#mkdir/1

lib/cadet/updater/CS1101S.ex
```diff
@@ -0,0 +1,69 @@
...
+
+  def clone() do
+    Logger.info("Cloning CS1101S: Started")
+    git("clone", [@remote_repo, @local_name])
+    Logger.info("Cloning CS1101S: Done")
+  end
+
+  def update() do
+    Logger.info("Updating CS1101S...")
+    git("fetch")
```
indocomsoft: Why do you need to `git fetch`? Afaik `git pull` = `git fetch` + `git merge`

indocomsoft:
@ningyuansg w.r.t. `.env` vs `config/prod.secret.exs`, personally I prefer the latter since https://github.com/avdi/dotenv_elixir clearly says in the header `WARNING: This isn't the Elixir way.`.

With that in mind, you're the deployment guy so far, so do what you feel is best. It'd be really great if we can move away from `dotenv` and embrace the full elixir way haha

lib/cadet/updater/CS1101S.ex
ningyuansg: Nope, it does something like `$ git clone git@github.com/cs1101s/cs1101s.git .` which creates a new directory `cs1101s` in cwd, which is the mix project root in dev/test and wherever you run the cadet binary from in prod.

lib/cadet/updater/CS1101S.ex
ningyuansg: You're 100% right here I just like to fetch before I pull so often that I forgot pull does an implicit fetch.Fixed with 69da032.

test/cadet/updater/CS1101S_test.exs
ningyuansg: `&` cannot create zero arity anonymous functions, I think.

test/support/data_case.ex
ningyuansg: Fixed with 7874796.

ningyuansg:
@indocomsoft Made the requested changes, I've replied on the threads w.r.t. changes not made (just the one about `clone/0` and the use of `&` vs `fn`). I'll make an issue for the dotenv thing.

test/cadet/updater/CS1101S_test.exs
indocomsoft: Ahh, right, my bad

lib/cadet/updater/CS1101S.ex
indocomsoft: Oh rightttt. Sorry I derped

ningyuansg:
@indocomsoft Coverage comes from the lazy debug calls, which do not get used since the test config only prints logs with warn level and above. Credo won't allow anything but lazy debug calls (when interpolation is used), apparently there are performance hits if you don't use them.

indocomsoft:
https://hexdocs.pm/logger/Logger.html#module-application-configuration

You can get elixir to "purges at compilation time all calls that have log level lower than the value of this option."

It does not make sense to compile logs lower than the level set in the config file. You should set in `config/test.exs`

```elixir
config :logger, level: :warn, compile_time_purge_level: :warn
```

And coverage shld be complete then since those `Logger.debug` lines aren't compiled \o/

indocomsoft:
@ningyuansg bump

ningyuansg:
Fixed with c6bbb5d

PR #122, opened 2018-07-15
Remove json deserialisers

indocomsoft:
Fixes #103

PR #123, opened 2018-07-15
Change Poison -> Jason for JSON parser and generator

indocomsoft:
For performance reasons: quoting its README:
> The parser and generator are at least twice as fast as other Elixir/Erlang libraries (most notably `Poison`). The performance is comparable to `jiffy`, which is implemented in C as a NIF. Jason is usually only twice as slow.

### DIFF
--- a/lib/cadet/accounts/ivle.ex
+++ b/lib/cadet/accounts/ivle.ex
@@ -129,7 +129,7 @@ defmodule Cadet.Accounts.IVLE do
   """
   def api_call(method, queries) when method in ["Announcements"] do
     with {:ok, %{status_code: 200, body: body}} <- HTTPoison.get(api_url(method, queries)),
-         body = Poison.decode!(body),
+         body = Jason.decode!(body),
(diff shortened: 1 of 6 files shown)


tuesmiddt:
Can you regenerate the lockfile as well?

indocomsoft:
Eh, I don't think I need to since `jason` is already a dependency of `excoveralls`

indocomsoft:
In any case, I recreated `mix.lock` and there's an update to `meck` apparently

tuesmiddt:
Ah I was thinking poison could be removed from the lockfile but I think something else depends on it. 

In any case, `Poison` is still being used in the following test files: 
- assessments_test
- question_test
- mcq_question_test
- programming_question_test

Do you want to change them as well?

indocomsoft:
@tuesmiddt actually there's #122 (removing json deserialisers) which would remove some changes from here, so I think you should look at that PR first before this one.

PR #125, opened 2018-07-16
Split factory into separate files

### LINKED ISSUE #93: Move factory.ex to test/support and separate them into different files
Refer to https://github.com/thoughtbot/ex_machina#splitting-factories-into-separate-files

indocomsoft:
Fixes #93


test/factories/accounts/user_factory.ex
```diff
@@ -0,0 +1,25 @@
...
+  """
+
+  defmacro __using__(_opts) do
+    quote do
+      alias Cadet.Accounts.User
+
+      def user_factory do
+        %User{
+          name: Faker.Name.En.name(),
+          role: :staff
```
tuesmiddt: Let's set this to random role? although it doesn't really make a difference since we usually specify a role when creating the user anyway.

indocomsoft: Done.

PR #131, opened 2018-07-16
WIP: Update user submission `updated_at` on answer submit

### LINKED ISSUE #127: "Betcha" button for missions/quests/paths
> The idea is to let students "bet" that their assessment solution is correct. Their solution would be irrevocably frozen, similar to an expiry of the assessment deadline. The students would be credited with XP, corresponding to the number of hours between pressing the button and the deadline.

Originally opened by Prof @martin-henz, at https://github.com/source-academy/cadet-frontend/issues/202. Moving here since implementation depends more on the backend.

tuesmiddt:
Fixes #127

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -7,6 +7,7 @@ defmodule Cadet.Assessments do
 
   import Ecto.Query
 
+  alias Ecto.Multi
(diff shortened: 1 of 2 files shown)


tuesmiddt:
Let me keep the branch there for reference for now. Delete after 4 Aug.

PR #116, opened 2018-07-14
Assessments controller

tuesmiddt:
Fixes #66,
Fixes #84 

Responses and params are in swagger schema.
These two endpoints are available for all users.

Assessments index endpoint returns `attempted: false` for all non-students since they cannot submit answers.

PR #133, opened 2018-07-17
Refactor assessments view

tuesmiddt:
Fixes #116


lib/cadet_web/views/view_helpers.ex
```diff
@@ -6,4 +6,77 @@ defmodule CadetWeb.ViewHelpers do
...
+    barbar: "ha",
+    foobar: "hoha"
+  }
+
+  field_list = [:foofoo, :barbar]
+
+  transform_map_for_view(source, field_list)
+  > %{
+    foofoo: "ho",
+    barbar: "bar"
```
indocomsoft: Did you mean `barbar: "ha"` lol

lib/cadet_web/views/view_helpers.ex
```diff
@@ -6,4 +6,77 @@ defmodule CadetWeb.ViewHelpers do
...
+  }
+
+  transform_map_for_view(source, key_transformations)
+  > %{
+    foo: Map.get(source, :foofoo),
+    bar: Map.get(source, :barbar)
+  }
+
+  function_transformations = %{
+    foo: fn source -> source.foofoo <> "hoho",
```
indocomsoft: I think this is a better design:
```elixir
source = %{
  foofoo: "ho",
  barbar: "ha",
  foobar: "hoha"
}

function_transformations = %{
  foofoo: &(&1 <> "tst")
}

transform_map_for_view(source, functions_transformations)
# > %{
#   foofoo: "hotst"
# }
```

Not really sure why you decided to do what you did. Maybe you could elaborate?

tuesmiddt: For me, the primary intention is to create a standardised way to transform from a `source_map` into a `view_map`. Other than the direct `source_field_name` to `view_field_name` transformation, I wanted to be able to define a `view_field` as a `fn source_map -> my_custom_transformation_to_value(source_map) end`. The transformation could involve aggregation of multiple fields from the `source_map`, calculations, etc. 

Your suggestion only applies to when `source_map` and `view_map` have the same `field_name` and you are doing a direct `value` -> `new_value` transformation.

indocomsoft: Ah ok fair enough.

PR #157, opened 2018-07-22
Redefine library per new schema

### LINKED ISSUE #149: Redefining Library
Recently, I've implemented the external libraries on the frontend, and in doing so understood a bit more about how the interpreter uses the external libraries. As such, a slight change in the structure of the `Library` (under `Question`) is required.

Right now the structure is [as such](https://github.com/source-academy/cadet-frontend/blob/97268b8d14bd0249ff98f6dd6e3cbe8215296917/src/components/assessment/assessmentShape.ts#L66).
- `chapter`
- `externals` is a list of strings that expose global functions to the interpreter (such as `show`, that renders a rune onto a webGL object).
- `files` is a list of strings used to load up libraries. This is not needed anymore as the website is static, meaning that all the libraries must be served initially and used when it is needed.
- `globals` is a list of strings. It has been unused so far, but I misinterpreted its usage before. Updated the usage below.

This is a structure that would [better](https://github.com/source-academy/cadet-frontend/blob/92a14499e3638ccbaa1bd16ad5e7363c7566a809/src/components/assessment/assessmentShape.ts) suit the needs of the website:
Library:
 -  `chapter: number` (no change)
  - `externalLibraryName: "NONE" | "TWO_DIM_RUNES" | "THREE_DIM_RUNES" | "CURVES" | "SOUND"`. This will be used to specify what external library (rendering webGL or playing sound) the assessment would be using. I considered inferring this from `externals`, but a field is more explicit and leaves no room for error on our part.
  - `externals: string[]` (no change)
  - `globals: <string, value>[]`. This is a list of pairs (`string`, `value`) that can be used to specify constants. This will be useful for the staff to introduce default values for the students or external libraries.

Do let me know if there are issues regarding data representation, @ningyuansg @indocomsoft @tuesmiddt

tuesmiddt:
Fixes #149 

@remo5000 @ningyuansg This PR changes `GET /grading/{submissionId}`'s `question` field slightly:
`questionId` -> `id`,
`questionType` -> `type`


lib/cadet_web/endpoint.ex
```diff
@@ -16,8 +16,8 @@ defmodule CadetWeb.Endpoint do
   # Code reloading can be explicitly enabled under the
   # :code_reloader configuration of your endpoint.
   if code_reloading? do
-    socket("/phoenix/live_reload/socket", Phoenix.LiveReloader.Socket)
-    plug(Phoenix.LiveReloader)
+    # socket("/phoenix/live_reload/socket", Phoenix.LiveReloader.Socket)
```
tuesmiddt: LMK if i can delete this

lib/cadet/assessments/library.ex
```diff
@@ -1,22 +1,25 @@
...
-  The library entity represents a library to be used in a programming question.
+  The library entity represents a library to be used in a  question.
   """
   use Cadet, :model
 
+  alias Cadet.Assessments.Library.ExternalLibrary
+
   embedded_schema do
     field(:chapter, :integer, default: 1)
     field(:globals, {:array, :string}, default: [])
```
indocomsoft: Hmm in the frontend, the type of `globals` is array of "tuple" of string and any (`Array<[string, any]>`). Maybe we can represent it as a `:map`.

lib/cadet_web/endpoint.ex
indocomsoft: Yeah, just delete these 2 lines.

lib/cadet/assessments/library.ex
tuesmiddt: ah yes i missed this

lib/cadet/assessments/library.ex
```diff
@@ -21,5 +21,16 @@ defmodule Cadet.Assessments.Library do
     |> cast(params, @required_fields ++ @optional_fields)
     |> cast_embed(:external)
     |> validate_required(@required_fields ++ @required_embeds)
+    |> validate_globals()
+  end
+
+  def validate_globals(changeset) do
+    globals = get_change(changeset, :globals)
+
+    if Enum.all?(globals, fn {name, value} -> is_binary(name) and is_binary(value) end) do
```
indocomsoft: Lol I think value can be anything hor
```typescript
const mockGlobals: Array<[string, any]> = [
  ['testNumber', 3.141592653589793],
  ['testString', 'who dat boi'],
  ['testBooleanTrue', true],
  ['testBooleanFalse', false],
  ['testBooleanUndefined', undefined],
  ['testBooleanNull', null],
  ['testObject', { a: 1, b: 2 }],
  ['testArray', [1, 2, 'a', 'b']]
]
```

ningyuansg:
@remo5000 Do take a look, I think you're more familiar with frontend requirements for this than I am

indocomsoft:
Based on the discussion in #149, `globals` is a dictionary of `string` to `string`.

PR #163, opened 2018-07-24
Fix changeset helper to use casted embedded struct

### LINKED ISSUE #162: MCQ answer returned as string instead of number
#### Steps to reproduce
- Authenticate
- Request for an Assessment with an answered mcq question

I'm getting this: 
image

tuesmiddt:
Fixes #162 

For context, `answer: %{choice_id: "1"}` can be validly cast with a `field(:choice_id, :integer)` so it passes validation, but this will not be reflected in the original changeset.

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -291,15 +291,18 @@ defmodule Cadet.Assessments do
   defp insert_or_update_answer(submission = %Submission{}, question = %Question{}, raw_answer) do
     answer_content = build_answer_content(raw_answer, question.type)
 
-    %Answer{}
-    |> Answer.changeset(%{
-      answer: answer_content,
-      question_id: question.id,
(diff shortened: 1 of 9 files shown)


lib/cadet/assessments/assessments.ex
```diff
@@ -291,15 +291,18 @@ defmodule Cadet.Assessments do
...
+      |> Answer.changeset(%{
+        answer: answer_content,
+        question_id: question.id,
+        submission_id: submission.id,
+        type: question.type
+      })
+
+    Repo.insert(
+      answer_changeset,
+      on_conflict: [set: [answer: get_change(answer_changeset, :answer)]],
```
indocomsoft: Not sure what's the difference compared to the previous one?

indocomsoft: Ahh ok I see.

PR #168, opened 2018-07-27
Add DeploymentHelper for selective compilation based on Mix.env()

indocomsoft:
Similar to the `#if` preprocessor in C.

Let me know if any of you have a better name for `if_compilation`


ningyuansg:
How bout `if_env` (this pr closed alr tho)

PR #170, opened 2018-07-28
Add story to /user endpoint

### LINKED ISSUE #148: Assessments need additional property/field to map to a story in the game
The game comes in two modes,

1. An open-world mode, where the user can explore the spaceship and talk to NPCs with no restrictions
2. A scripted mode ('*story*'), where the user is guided through conversations with specific NPCs in specific locations

There are many stories, most ending in a link that goes to a certain assessment. Therefore, each story maps to a certain assessment. In #84, we added the capability to ignore stories mapping to assessments that have been attempted. If all stories are ignored, the open-world mode is rendered for the user.

---

The problem now is that we don't have a way to map a story to an assessment. The current `id` field of an assessment is only it's insertion order into the database. Consider the following functions from [source-academy](https://github.com/source-academy/source-academy2/blob/11466e86facfdcb3527731af0e914baa1da845ef/lib/source_academy/assessments.ex#L642),

```elixir
  def prepare_game(student) do
    # Get all non-attempted assessment
    assessments = Repo.all(
      from a in Assessment,
        left_join: s in Submission,
          on: a.id == s.assessment_id and s.student_id == ^student.id,
        where: is_nil(s.student_id),
        select: a
    )
    # Filter those can be attempted, min by order, mission first
    assessments =
      assessments
      |> Enum.filter(&(&1.type != :path && can_attempt?(&1, student.user)))
    if Enum.empty?(assessments) do
      nil
    else
      [hd | _ ] = Enum.sort_by(assessments,
        &story_name_pair/1,
        &compare_story_name_pair/2)
      hd
    end
  end

  def story_name_pair(assessment) do
    {assessment.type, assessment.name}
  end

  def compare_story_name_pair({t1, n1}, {t2, n2}) do
    cond do
      t1 == t2 -> n1 <= n2
      t1 == :mission -> true
      t1 == :sidequest && t2 == :mission -> false
      t1 == :contest && t2 == :mission -> false
      t1 == :contest && t2 == :sidequest -> false
      true -> false
    end
  end

  def story_name(assessment) do
    type = Atom.to_string(assessment.type)
    String.downcase(type) <> "-" <> assessment.name
  end
```

In the above, assessments are sorted according to the comparison in `compare_story_name_pair/2` and filtered at `prepare_game/1` student. A corresponding string representing the story to show in the game is then generated using `story_name/1`. 

The 'story_names' all correspond to the filename of an XML file from `./stories/` in the cs1101s repository. We pass a *'story_name'* to the javascript function starting the game, and it will fetch the appropriate XML file and assets from AWS S3, and load them up for the user ([the JS function](https://github.com/source-academy/cadet-frontend/blob/97268b8d14bd0249ff98f6dd6e3cbe8215296917/src/components/academy/game/create-initializer.js#L72)).

---

Overall, two problems to solve:

- There is a need for an addition property/field, perhaps *'story_name'*, which can be compared, and sorted just like in source-academy2 to determine the load order of stories
    - This new property, representing which story to load should then be added to the `/user` endpoint
- Since there is a change in the names of missions in the cs1101s repository, there is a need to either:
    - Rename the story XML files in a similar fashion, or 
    - Transform the new mission filenames to current story XML filenames when sending it to the frontend

### LINKED ISSUE #161: Add fields/entities to schema based on XML spec
- `Question`?
  - [ ] `Library` for the auto-grader (`GRADERDEPLOYMENT`)
- `Assessment`
  - [ ] `number` containing e.g. `M06` (in contrast to title containing, e.g. `Gosperize`)
  - [ ] `story` containing the story name in the game for front-end
  - [ ] `READING` (e.g. Textbook Section 1.1.1 to 1.1.4)

tuesmiddt:
Fixes #161 
Fixes #148 

Needs a `mix ecto.reset` due to reordering of enums

### DIFF
--- a/test/cadet_web/controllers/user_controller_test.exs
+++ b/test/cadet_web/controllers/user_controller_test.exs
@@ -76,7 +76,9 @@ defmodule CadetWeb.UserControllerTest do
       late_assessments = build_assessments_starting_at(Timex.shift(Timex.now(), days: 1))
       assessments_with_indices = Enum.with_index(early_assessments ++ late_assessments)
 
-      for {assessment, idx} <- Enum.slice(assessments_with_indices, 0, 7) do
+      # Submit for i-th assessment, expect (i+1)th story to be returned
+      for {assessment, idx} <-
+            Enum.slice(assessments_with_indices, 0, length(assessments_with_indices) - 1) do
         insert(:submission, %{student: user, assessment: assessment})
 
         {expected_assessment, _} = Enum.fetch!(assessments_with_indices, idx + 1)
--- a/test/cadet_web/controllers/user_controller_test.exs
+++ b/test/cadet_web/controllers/user_controller_test.exs
@@ -77,8 +77,7 @@ defmodule CadetWeb.UserControllerTest do
       assessments_with_indices = Enum.with_index(early_assessments ++ late_assessments)
 
       # Submit for i-th assessment, expect (i+1)th story to be returned
-      for {assessment, idx} <-
-            Enum.slice(assessments_with_indices, 0, length(assessments_with_indices) - 1) do
+      for {assessment, idx} <- Enum.drop(assessments_with_indices, -1) do
         insert(:submission, %{student: user, assessment: assessment})
 
         {expected_assessment, _} = Enum.fetch!(assessments_with_indices, idx + 1)
--- a/test/cadet_web/controllers/user_controller_test.exs
+++ b/test/cadet_web/controllers/user_controller_test.exs
@@ -3,7 +3,9 @@ defmodule CadetWeb.UserControllerTest do
 
   import Cadet.Factory
 
+  alias Cadet.Repo
   alias CadetWeb.UserController
+  alias Cadet.Assessments.{Assessment, AssessmentType}
 
   test "swagger" do
     assert is_map(UserController.swagger_definitions())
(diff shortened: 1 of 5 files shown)


test/cadet_web/controllers/user_controller_test.exs
```diff
@@ -12,31 +14,120 @@ defmodule CadetWeb.UserControllerTest do
...
+
+    @tag authenticate: :student
+    test "success, student story skips submitted", %{conn: conn} do
+      user = conn.assigns.current_user
+
+      early_assessments = build_assessments_starting_at(Timex.now())
+      late_assessments = build_assessments_starting_at(Timex.shift(Timex.now(), days: 1))
+      assessments_with_indices = Enum.with_index(early_assessments ++ late_assessments)
+
+      for {assessment, idx} <- Enum.slice(assessments_with_indices, 0, 7) do
```
indocomsoft: Lol magic number `7` where from

test/cadet_web/controllers/user_controller_test.exs
```diff
@@ -12,31 +14,122 @@ defmodule CadetWeb.UserControllerTest do
...
+    test "success, student story skips submitted", %{conn: conn} do
+      user = conn.assigns.current_user
+
+      early_assessments = build_assessments_starting_at(Timex.now())
+      late_assessments = build_assessments_starting_at(Timex.shift(Timex.now(), days: 1))
+      assessments_with_indices = Enum.with_index(early_assessments ++ late_assessments)
+
+      # Submit for i-th assessment, expect (i+1)th story to be returned
+      for {assessment, idx} <-
+            Enum.slice(assessments_with_indices, 0, length(assessments_with_indices) - 1) do
```
indocomsoft: I think `Enum.drop(assessments_with_indices, -1)` conveys the intent clearer

indocomsoft:
Also could you merge from/rebase on `master`

test/cadet_web/controllers/user_controller_test.exs
tuesmiddt: Done

tuesmiddt:
and done

PR #171, opened 2018-07-28
Add additional grading info

### LINKED ISSUE #165: Add assessment name to "/grading" endpoint result
I feel that staff would recognise assessments using their name, and the `assessmentId` would be hard to keep track of. I've included it in the frontend since making the grading listings, and it looks pretty intuitive on the table to see the assessment name

tuesmiddt:
Fixes #165


ningyuansg:
@remo5000 Please note change in field names w.r.t. JSON response.

PR #172, opened 2018-07-28
[WIP] Ivle integration

sreycodes:
1. A certain Postgres error comes up when I try to create announcements:
test Send :work to GenServer Invalid token (Cadet.Updater.PublicTest)
     test/cadet/updater/public_test.exs:108
     ** (DBConnection.OwnershipError) cannot find ownership process for #PID<0.868.0>.
     
     When using ownership, you must manage connections in one
     of the four ways:
     
     * By explicitly checking out a connection
     * By explicitly allowing a spawned process
     * By running the pool in shared mode
     * By using :caller option with allowed process
     
     The first two options require every new process to explicitly
     check a connection out or be allowed by calling checkout or
     allow respectively.
     
     The third option requires a {:shared, pid} mode to be set.
     If using shared mode in tests, make sure your tests are not
     async.
     
     The fourth option requires [caller: pid] to be used when
     checking out a connection from the pool. The caller process
     should already be allowed on a connection.
     
     If you are reading this error, it means you have not done one
     of the steps above or that the owner process has crashed.
     
     See Ecto.Adapters.SQL.Sandbox docs for more information.
     code: assert {:noreply, new_api_params} = Public.handle_info(:work, api_params)
     stacktrace:
       (db_connection) lib/db_connection.ex:934: DBConnection.checkout/2
       (db_connection) lib/db_connection.ex:750: DBConnection.run/3
       (db_connection) lib/db_connection.ex:592: DBConnection.prepare_execute/4
       (ecto) lib/ecto/adapters/postgres/connection.ex:73: Ecto.Adapters.Postgres.Connection.prepare_execute/5
       (ecto) lib/ecto/adapters/sql.ex:256: Ecto.Adapters.SQL.sql_call/6
       (ecto) lib/ecto/adapters/sql.ex:426: Ecto.Adapters.SQL.execute_and_cache/7
       (ecto) lib/ecto/repo/queryable.ex:133: Ecto.Repo.Queryable.execute/5
       (ecto) lib/ecto/repo/queryable.ex:37: Ecto.Repo.Queryable.all/4
       (ecto) lib/ecto/repo/queryable.ex:70: Ecto.Repo.Queryable.one/4
       (cadet) lib/cadet/updater/public.ex:73: anonymous fn/1 in Cadet.Updater.Public.handle_info/2
       (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
       (cadet) lib/cadet/updater/public.ex:71: Cadet.Updater.Public.handle_info/2
       test/cadet/updater/public_test.exs:111: (test)

2. To make sure there are no duplicate announcements, there are two announcements:
a. Either we decide on a way to say that 2 announcements are same (something like checking for title and description) OR
b. The IVLE API provides an isRead flag which becomes true when the certain user logs in and reads the announcement (not when the API is called). I was having some trouble simulating such behaviour and need some help if we decide to go on this route.

3. Regarding pulling data about the groups, the IVLE API seems to not be working properly. Therefore we have to come up with some other way to update the "Groups" database.


config/config.exs
```diff
@@ -53,3 +53,13 @@ config :guardian, Guardian.DB,
   token_types: ["access"],
   # default: 60 minute
   sweep_interval: 60
+
+# Configure :ex_aws
+config :ex_aws,
+  access_key_id: {:system, "AWS_ACCESS_KEY_ID"},
+  secret_access_key: {:system, "AWS_SECRET_ACCESS_KEY"},
```
ningyuansg: Add `:instance_role` so that in production, `ex_aws` will look for instance role credentials in our ec2 instances. Otherwise, this current configuration will fail in production.

config/config.exs
```diff
@@ -53,3 +53,13 @@ config :guardian, Guardian.DB,
...
+  s3: [
+    scheme: "https://",
+    host: "sreyansapitest.s3-ap-southeast-1.amazonaws.com",
```
ningyuansg: Please add a corresponding example in `prod.secrets.exs` as a template for production.

tuesmiddt:
1. You have to have `use Cadet.DataCase` inside your test file because it adds a helper that checks out a database connection per test. Next time, you might want to read the documentation that the error points you to. `See Ecto.Adapters.SQL.Sandbox docs for more information.`
2. What is the response that IVLE sends you? You should mock the responses in `cadet/test/fixtures` for your tests too.
3. What do you suggest we do instead?

PR #173, opened 2018-07-28
Betcha functionality

### LINKED ISSUE #127: "Betcha" button for missions/quests/paths

### LINKED ISSUE #178: Swagger UI out of date for GET /user
No field `story` is shown. Off e1bda9.

tuesmiddt:
Fixes #127 
Fixes #178

@ningyuansg @remo5000 
Outline of changes:
- New submission status "not_attempted/attempting/attempted/submitted" instead of old attempted flag
- New endpoint `POST /assessments/id/submit` to finalise submission if current status is `attempted` (see tests/swagger for more details)
- When a student answers a question, we will check whether every question in the assessment has been answered. If yes, submission status will be updated to `attempted`
- `story` field for `GET is chosen based on the following criteria:
  - Filter by:
    - assessment is published
    - assessment story is not nil
    - assessment `open_at` is before now and `close_at` is after now
    - submission for this question does not exist or has status `attempting`
  - Sort by:
    - Assessment `open_at`
    - Assessment type (path>mission>sidequest>contest) for assessments that open at the same time


ningyuansg:
> When a student answers a question, we will check whether every question in the assessment has been answered. If yes, submission status will be updated to `submitted`

Can I verify if this is meant to be '...updated to `attempted`' instead of `submitted`? 

If yes, `submitted` assessments are those that are past the deadline, or those that have received a `POST /assessment/id/submit`?

Then, `unattempted` are assessments which have never received an answer for any of their questions, and `attempting` are assessments with a mix of questions that have and have not received answers?

tuesmiddt:
1. Yes it should be `attempted`. My bad.
2. Submitted are those that have received a `POST /assessment/id/submit`.
3. Yep that is right, although it's currently called `not_attempted`. This is a fairly trivial change to make.

ningyuansg:
`not_attempted` is fine as well, that was a typo on my part. Either way we'll create the enum on our side based on this PR.

ningyuansg:
Thanks for the quick reply too!

test/cadet_web/controllers/user_controller_test.exs
```diff
@@ -69,32 +70,77 @@ defmodule CadetWeb.UserControllerTest do
...
-      late_assessments = build_assessments_starting_at(Timex.shift(Timex.now(), days: 1))
+      early_assessments = build_assessments_starting_at(Timex.shift(Timex.now(), days: -3))
+      late_assessments = build_assessments_starting_at(Timex.shift(Timex.now(), days: -1))
       assessments_with_indices = Enum.with_index(early_assessments ++ late_assessments)
 
       # Submit for i-th assessment, expect (i+1)th story to be returned
-      for {assessment, idx} <- Enum.drop(assessments_with_indices, -1) do
-        insert(:submission, %{student: user, assessment: assessment})
+      for status <- [:attempted, :submitted] do
+        for {assessment, idx} <- Enum.drop(assessments_with_indices, -1) do
```
indocomsoft: I think this can be more elegantly achieved using `Enum.chunk_every/2` which is comparable to Ruby's `Enumerable#each_cons`

test/cadet_web/controllers/user_controller_test.exs
```diff
@@ -69,32 +70,77 @@ defmodule CadetWeb.UserControllerTest do
...
+            conn
+            |> get("/v1/user")
+            |> json_response(200)
+            |> Map.get("story")
+
+          assert resp_story == expected_assessment.story
+        end
 
-        assert resp_story == expected_assessment.story
+        Repo.delete_all(Submission)
```
indocomsoft: 🙃

test/cadet_web/controllers/user_controller_test.exs
```diff
@@ -69,32 +70,77 @@ defmodule CadetWeb.UserControllerTest do
...
       end
     end
 
     @tag authenticate: :staff
     test "success, staff", %{conn: conn} do
       user = conn.assigns.current_user
+      build_assessments_starting_at(Timex.shift(Timex.now(), days: -3))
```
indocomsoft: Why do you need this? Haha

test/cadet_web/controllers/user_controller_test.exs
tuesmiddt: You're right. This is much nicer.

test/cadet_web/controllers/user_controller_test.exs
tuesmiddt: There's a unique constraint so submissions created for previous status need to be deleted. Can't delete in the inner loop or it breaks the test. ExUnit checks out a transaction for each test and all changes are rolled back after the test is complete so this shouldn't be an issue.

test/cadet_web/controllers/user_controller_test.exs
tuesmiddt: I initially forgot to ensure staff only get `nil` so I added this when I remembered.

tuesmiddt:
Fixed according to your comments.

PR #182, opened 2018-08-01
Add fallback user story fetching logic

### LINKED ISSUE #179: Need additional field `attempted_all` for GET /user
### Specification

- `attempted_all` is a new field in `GET /user`
- `attempted_all` is `true` if all assessments in `GET /assessments` has an `attempted` field of value attempting, attempted, or submitted. Otherwise (if there is a single field that is `unattempted`) it is `false`.
- The `story` field of `GET /user` must refer to a valid story XML file (it cannot be empty or null)
    - Then, if no valid assessments remain after the filter step in #173, please send us the story of the most recent assessment (that has a story)
    - `story` field of an assessment via `GET /assessments` or `GET /assessment/{id}` can be empty (we actually don't use this at all at frontend)

### Rationale

- The game is only rendered once. Basically, we don't know how to render it more than once (it was not written for SPAs). So we only need the `story` field from `GET /user`. The `story` field of the rest don't really matter, but no harm having it there.
- The game _absolutely_ requires a valid `story`, which corresponds to a story XML file, in order to start properly. 
    - This is because each `story` indicates which version of the story assets to load, e.g. deathcube stories have assets indicating that the spaceship's alarms are going off.
- The game takes an additional `attemptedAll` boolean parameter.
    - If attemptedAll is `true`, the story (i.e. all the dialogues and scripted events) will play
    - Otherwise, it is open-world

tuesmiddt:
Fixes #179 

When logic in #173 causes no story to be found, `story["allAttemped"]` will be `true` and the following logic will be used to find a story:
  - Filter by:
    - assessment is published
    - assessment story is not nil
    - assessment `open_at` is before now and `close_at` is after now
    - **NO** submission status filtering
  - Sort by:
    - Assessment `close_at`, descending
    - Assessment type (path>mission>sidequest>contest) for assessments that close at the same time

@ningyuansg Note that the `story` field may be `{story: null, allAttempted: true}` in the event that both methods fail to find a qualifying story. At the moment, the only time I see this happening is at the start of semester before any assessment is open.

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -29,20 +29,45 @@ defmodule Cadet.Assessments do
     end
   end
 
-  def user_current_story(%User{id: user_id, role: role}) do
-    if role in @submit_answer_roles do
-      Assessment
-      |> where(is_published: true)
-      |> where([a], not is_nil(a.story))
-      |> where([a], a.open_at <= from_now(0, "second") and a.close_at >= from_now(0, "second"))
-      |> join(:left, [a], s in Submission, s.assessment_id == a.id and s.student_id == ^user_id)
-      |> where([_, s], is_nil(s.id) or s.status == "attempting")
-      |> order_by([a], a.open_at)
-      |> order_by([a], a.type)
-      |> select([a], a.story)
-      |> first()
-      |> Repo.one()
+  def user_current_story(user = %User{}) do
+    {:ok, %{result: story}} =
+      Multi.new()
+      |> Multi.run(:default, fn _ -> {:ok, get_user_story_by_direction(user)} end)
+      |> Multi.run(:result, fn %{default: default_res} ->
+        if default_res,
+          do: {:ok, %{all_attempted: false, story: default_res}},
+          else: {:ok, %{all_attempted: true, story: get_user_story_by_direction(user, :backward)}}
+      end)
+      |> Repo.transaction()
+
+    story
+  end
+
+  @spec get_user_story_by_direction(%User{}, :forward | :backward) :: String.t() | nil
+  def get_user_story_by_direction(%User{id: user_id}, direction \\ :forward)
+      when is_atom(direction) do
+    filter_and_sort = fn query ->
+      case direction do
+        :forward ->
+          query
+          |> where([_, s], is_nil(s.id) or s.status == "attempting")
+          |> order_by([a], asc: a.open_at)
+
+        :backward ->
+          query |> order_by([a], desc: a.close_at)
+      end
     end
+
+    Assessment
+    |> where(is_published: true)
+    |> where([a], not is_nil(a.story))
+    |> where([a], a.open_at <= from_now(0, "second") and a.close_at >= from_now(0, "second"))
+    |> join(:left, [a], s in Submission, s.assessment_id == a.id and s.student_id == ^user_id)
+    |> filter_and_sort.()
+    |> order_by([a], a.type)
+    |> select([a], a.story)
+    |> first()
+    |> Repo.one()
   end
 
   def assessment_with_questions_and_answers(id, user = %User{}) when is_ecto_id(id) do
--- a/test/cadet_web/controllers/user_controller_test.exs
+++ b/test/cadet_web/controllers/user_controller_test.exs
@@ -39,13 +39,18 @@ defmodule CadetWeb.UserControllerTest do
       late_assessments = build_assessments_starting_at(Timex.shift(Timex.now(), days: -1))
 
       for assessment <- early_assessments ++ late_assessments do
-        story =
+        resp_story =
           conn
           |> get("/v1/user")
           |> json_response(200)
           |> Map.get("story")
 
-        assert story == assessment.story
+        expected_story = %{
+          "story" => assessment.story,
+          "allAttempted" => false
+        }
+
+        assert resp_story == expected_story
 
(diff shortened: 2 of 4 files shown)


lib/cadet/assessments/assessments.ex
```diff
@@ -29,20 +29,45 @@ defmodule Cadet.Assessments do
...
+  end
+
+  @spec get_user_story_by_direction(%User{}, :forward | :backward) :: String.t() | nil
+  def get_user_story_by_direction(%User{id: user_id}, direction \\ :forward)
+      when is_atom(direction) do
+    filter_and_sort = fn query ->
+      case direction do
+        :forward ->
+          query
+          |> where([_, s], is_nil(s.id) or s.status == "attempting")
```
indocomsoft: Prefer atom `:attempting`

lib/cadet/assessments/assessments.ex
```diff
@@ -29,20 +29,45 @@ defmodule Cadet.Assessments do
...
-      |> select([a], a.story)
-      |> first()
-      |> Repo.one()
+  def user_current_story(user = %User{}) do
+    {:ok, %{result: story}} =
+      Multi.new()
+      |> Multi.run(:default, fn _ -> {:ok, get_user_story_by_direction(user)} end)
+      |> Multi.run(:result, fn %{default: default_res} ->
+        if default_res,
+          do: {:ok, %{all_attempted: false, story: default_res}},
```
indocomsoft: If it ends up multi-line, you should use the syntactic sugar
```elixir
if default_res do
  {:ok, %{all_attempted: false, story: default_res}}
else
  {:ok, %{all_attempted: true, story: get_user_story_by_direction(user, :backward)}}
end

File: lib/cadet/assessments/assessments.ex [hunk 3]
```diff
@@ -29,20 +29,45 @@ defmodule Cadet.Assessments do
...
+          else: {:ok, %{all_attempted: true, story: get_user_story_by_direction(user, :backward)}}
+      end)
+      |> Repo.transaction()
+
+    story
+  end
+
+  @spec get_user_story_by_direction(%User{}, :forward | :backward) :: String.t() | nil
```
indocomsoft: Tbh I feel like `:forward` isn't a good enough description since you add another filtering as compared to `:backward`

lib/cadet/assessments/assessments.ex
indocomsoft: Also, if I read #179 correctly, `all_attempted === false` if and only if there is any open assessment whose submission's status is `unattempted` (which in our case is basically the `is_nil(s.id)`). I don't think `:attempting` should be here?

test/cadet_web/controllers/user_controller_test.exs
```diff
@@ -60,13 +65,18 @@ defmodule CadetWeb.UserControllerTest do
...
-      story =
+      resp_story =
         conn
         |> get("/v1/user")
         |> json_response(200)
         |> Map.get("story")
 
-      assert story == Enum.fetch!(assessments, 1).story
+      expected_story = %{
+        "story" => Enum.fetch!(assessments, 1).story,
```
indocomsoft: Ehh, `assessments` is a tuple right? Then I think you should pattern match on it instead.

lib/cadet/assessments/assessments.ex
tuesmiddt: 1. Atoms don't work here. I think the `where` macro just expands it into some raw SQL query (so ecto_enum library is not involved). All our enums as of now should be queried by their string value.

2. In the call, he said that the `all_attempted` flag is essentially to tell whether the first or second kind of logic is used. I personally feel like `attempting` should be treated the same as `not_attempted` in the sense that the assessment has not been "completed".

@ningyuansg Can you confirm whether `allAttempted` flag should be true or false when status is `attempting`?

test/cadet_web/controllers/user_controller_test.exs
tuesmiddt: It's a list and i'm getting the second item so I feel it actually looks worse:

```
      [_, assessment | _] = assessments

      expected_story = %{
        "story" => assessment.story,
        "allAttempted" => false
      }
```

lib/cadet/assessments/assessments.ex
tuesmiddt: Renamed it. Is this better?

lib/cadet/assessments/assessments.ex
tuesmiddt: Fixed

test/cadet_web/controllers/user_controller_test.exs
indocomsoft: Oops I derped

indocomsoft: Oh shoot sorry I derped

lib/cadet/assessments/assessments.ex
tuesmiddt: Clarified with him. It should be `false`.

indocomsoft: Alright, cool.

tuesmiddt:
@indocomsoft Fixed business logic and updated tests. Please review again.

@ningyuansg I renamed `allAttempted` to `playStory` with reversed values from before i.e. story belongs to `not_attempted` assessment -> `playStory: true` and vice versa for clarity

PR #195, opened 2018-08-09
Allow non-student to open all assessments, even those not open yet

### LINKED ISSUE #194: Assessments that aren't open yet is sent

indocomsoft:
Fixes #194 

@ningyuansg this also allows staffs using the docker image to see all assessments for testing.

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -13,6 +13,7 @@ defmodule Cadet.Assessments do
 
   @submit_answer_roles ~w(student)a
   @grading_roles ~w(staff)a
+  @open_all_assessment_roles ~w(staff admin)a
 
   def user_total_grade(%User{id: user_id}) do
     grade =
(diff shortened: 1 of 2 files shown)


tuesmiddt:
Actually hold on, the previous behaviour was to send all assessment overviews as long as they're published. Filtering by `open_at` was only done for the `GET /assessments/id` endpoint. Was there a request to change this behaviour?

indocomsoft:
@tuesmiddt Oh actually it was me thinking it doesn't make sense to send them yet-to-be-open assessments? Or at least this was the behaviour of the old source academy. But now that I think more about it again, the behaviour you implemented makes more sense.

Maybe @ningyuansg and @remo5000 can chime in?

Currently the XML file is also not providing `is_published` for assessment, so my xml parser is just giving `is_published: true`. We can keep the current behaviour of making `open_at` affect only the `GET /assessments/id`, but maybe somehow the XML file should provide a way to specify `is_published` to hide assessments from students. And then I can adjust such that `is_published` only affects visibility for students.

tuesmiddt:
I think another concern is how the creation of assessments will be paced. If avengers are going to continue working on editing the XMLs for the new schema, it might be better to have your behaviour so the students don't see how WIP the course material is.

indocomsoft:
If the XML file provides information about `is_published` attribute, then WIP assessments can be marked as not published which will be hidden.

remo5000:
I'm for this idea as well, as long as the security is solid. I think it's a pretty good feeling to test out an assessment in prod before it's out

tuesmiddt:
I was more concerned with students seeing the list of assessments grow hahah. Anyway I realised that you didn't change the other function (assessments with questions and answers) that's necessary for staff to actually view the assessment details. sorry I missed that just now in my review.

ningyuansg:
Can we also, in this PR, allow all roles to submit answers to assessments?

tuesmiddt:
@ningyuansg Could you please reply here instead of via email? It appears that email responses are not added to the comment chain.

> Can we also, in this PR, allow all roles to submit answers to assessments?

I'm not in favour of this. The only benefit I see is allowing easier testing by staff, which should be done via test accounts on the staging environment instead.

martin-henz:
OK, Shaowei, you suggest we use a guest account for testing? We have created two guest accounts, gst77777 and gst88888, and could give all Avengers the password, for testing in the staging environment. Would that be your suggestion?

tuesmiddt:
Yes, I think that would be a better idea.

ningyuansg:
@tuesmiddt My comment showed up on email but not here because I deleted it quite soon after I posted it. My bad.

indocomsoft:
Updated. So to recap, my PR is changing the behaviour such that:
- `GET /assessments` returns all published assessment regardless of `open_at`
- `GET /assessments/{id}` returns detailed information about an assessment and its questions:
  - When the current time is after `open_at` for students
  - Whenever for `admin` and `staff`

tuesmiddt:
LGTM

PR #200, opened 2018-08-09
Add comment field to the response of AssessmentsController.show

### LINKED ISSUE #192: Comments not passed to Assessments
This is required to show students their comments, if there any. I propose a `comments` field that is `null` if there are no comments, or a `string`, under the `/assessments/{assessmentId}` endpoint.

indocomsoft:
Fixes #192

### DIFF
--- a/lib/cadet_web/views/assessments_view.ex
+++ b/lib/cadet_web/views/assessments_view.ex
@@ -73,7 +73,7 @@ defmodule CadetWeb.AssessmentsView do
        }) do
     components = [
       build_question(%{question: question}),
-      build_answer_and_comment_by_question_type(%{question: question}),
+      build_answer_fields_by_question_type(%{question: question}),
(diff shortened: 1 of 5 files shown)


lib/cadet_web/views/assessments_view.ex
```diff
@@ -73,7 +73,7 @@ defmodule CadetWeb.AssessmentsView do
        }) do
     components = [
       build_question(%{question: question}),
-      build_answer_by_question_type(%{question: question}),
+      build_answer_and_comment_by_question_type(%{question: question}),
```
tuesmiddt: What do you think about renaming this to just `build_answer_fields_by_question_type`? I foresee us adding more to this in future.

indocomsoft: Makes sense. Done

PR #203, opened 2018-08-11
Add timeout to lambda, fix name conflicts

ningyuansg:
1. Lambda timeouts (as well as the grader timeout) has been added as part of the terraform files.
    - Lambda timeout is the duration before AWS timeouts the node process; a lambda timeout results in a 504 Gateway Timeout, and all information is lost.
    - Grader timeout is the duration that the grader program gives each student + grader program combination to run (they run in parallel); a grader timeout results in a error only for that one student + grader program, and the other successful programs are unaffected (they still return their grades). See https://github.com/source-academy/grader/pull/15.
2. Fix some problems with name conflicts between staging and production resources.

### DIFF
--- a/terraform/cadet/inputs.tf
+++ b/terraform/cadet/inputs.tf
@@ -34,6 +34,14 @@ variable "lambda_filename" {
   description = "Location of the lambda deployment zip"
 }
 
+variable "lambda_timeout" {
+  description = "Timeout duration of the lambda service (seconds)"
+}
+
+variable "grader_timeout" {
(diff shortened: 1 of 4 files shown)


indocomsoft:
@tuesmiddt Not sure if the timeout case is considered in your #188

tuesmiddt:
@indocomsoft it's covered here https://github.com/source-academy/cadet/blob/b22421a2dc20a91a9be0c9732972530a787211a5/lib/cadet/jobs/autograder/lambda_worker.ex#L24

indocomsoft:
Ah OK cool

PR #204, opened 2018-08-12
Add sentry

tuesmiddt:
Adds sentry.

Need to change build script to run `mix sentry_recompile` instead of `mix compile` so that sentry source code context will have correct version of code. Not sure how to do that, maybe @ningyuansg can help?

Might need some additional changes to dockerfile per https://blog.lelonek.me/how-to-use-sentry-io-with-elixir-in-docker-8374e695f706. @indocomsoft help me out here.

I'll create new gmail and sentry accounts and post details in slack. Do not merge/redeploy before that.


ningyuansg:
All I can contribute is that the deployment packages are built with distillery's `MIX_ENV=prod mix release --env=prod`, and that the configuration for this process is at [`rel/config.exs`](https://github.com/source-academy/cadet/blob/master/rel/config.exs).

The rest may be up to trial-and-error with the staging server.

indocomsoft:
@ningyuansg is there any way to test distillery locally?

ningyuansg:
I haven't had to explore that option, so can't really say, sorry. I think staging is an acceptable testing site though.

indocomsoft:
@tuesmiddt if I understand it correctly, `mix sentry_recompile` is basically forcing `mix` to recompile sentry right? If so I feel that this shouldn't be an issue with deployment since it seems like everything is compiled from scratch.

@ningyuansg is it possible to deploy this branch to staging for @tuesmiddt to test out?

indocomsoft:
My bad. @ningyuansg all you have to do is, before running `MIX_ENV=prod mix release --env=prod`, you simply have to run `MIX_ENV=prod mix sentry_recompile` first.

tuesmiddt:
Checked on staging. Works. Staging `secrets.exs` is already set correctly (reports errors to "cadet_staging") project in sentry. Production `secrets.exs` must be configured before compiling for deployment.

indocomsoft:
Testing docker...

indocomsoft:
Alright, docker works. Merging this.

PR #214, opened 2018-08-15
Add info log when start autograding

indocomsoft:


indocomsoft:
As discussed, since the jobs are async, there's no way to check whether it's done. Might add a check whether there's any job left after like 3 hours or sth

PR #236, opened 2018-08-20
Refactor git in CS1101S for better error reporting

indocomsoft:


lib/cadet/jobs/updater/CS1101S.ex
```diff
@@ -42,34 +42,25 @@ defmodule Cadet.Updater.CS1101S do
...
-    Logger.debug(fn ->
-      "git #{cmd} #{inspect(args)} exited with #{exit}: #{out}"
-    end)
+    if exit != 0 do
+      error_message = "git #{cmd} #{inspect(args)} errored with exit code #{exit}\n#{out}"
+      Logger.error(error_message)
+      Sentry.capture_message(error_message)
+    else
+      Logger.info("git #{cmd} #{inspect(args)} succeeded")
+      Logger.debug(fn -> "\n#{out}" end)
```
tuesmiddt: I don't quite understand why this is wrapped in a function

indocomsoft: Oh, it's from credo basically it's recommended for `Logger.debug`'s argument to be evaluated lazily because it might contain something expensive, and sometimes in the config, `:logger` is set to not emit `debug` logs.

tuesmiddt: Oh otherwise it won't be lazily evaluated even if the log level is set to not report it. Thanks for the explanation!

ningyuansg: Elixir 1.7 fixes this: https://elixir-lang.org/blog/2018/07/25/elixir-v1-7-0-released/#logger-compile-time-purging

PR #237, opened 2018-08-20
Regard empty directory or no xml file as :ok

indocomsoft:


lib/cadet/jobs/xml_parser.ex
```diff
@@ -55,8 +55,8 @@ defmodule Cadet.Updater.XMLParser do
       {:assessment_type, false} -> {:error, "XML location of assessment type is not defined."}
       {:cloned?, _} -> {:error, "Local copy of repository is either missing or empty."}
       {:type, false} -> {:error, "Directory containing XML is not found."}
-      {:listing, _} -> {:error, "Directory containing XML is empty."}
-      {:filter, _} -> {:error, "No XML file is found."}
+      {:listing, _} -> :ok
```
tuesmiddt: If you want to keep this behaviour, It doesn't quite make sense to separate those 2 steps like this. Alternatively, you could `Logger.info()` the previous error message.

indocomsoft: Hmm, you're right. I'll do `Logger.info()` instead

indocomsoft:
@tuesmiddt done

PR #243, opened 2018-08-24
Allow admin to see and edit all submissions

### LINKED ISSUE #242: Allow `:admin` users to see all submmissions

indocomsoft:
Fixes #242


tuesmiddt:
I get why you're not exactly happy about this. Our permission management seems kind of arbitrary and hard-coded. I think it's fine for now due to time constraints and since we don't foresee many additions to it. A refactor might be in order if we need to add more granularity.

PR #255, opened 2018-08-31
Allow for Batch Grading Process

rrtheonlyone:
Addresses Issues #254 and #253 

`/group` endpoint added to show all students under a grader
`/grading` endpoints modified to allow all submissions to be viewed and graded

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -16,7 +16,7 @@ defmodule Cadet.Assessments do
   @xp_bonus_assessment_type ~w(mission sidequest)a
   @submit_answer_roles ~w(student)a
   @grading_roles ~w()a
-  @see_all_submissions_roles [:staff, :admin]
+  @see_all_submissions_roles ~w(staff admin)a
   @open_all_assessment_roles ~w(staff admin)a
 
   @spec user_total_xp(%User{}) :: integer()
@@ -524,44 +524,51 @@ defmodule Cadet.Assessments do
         attrs,
         grader = %User{role: role}
       )
-      when is_ecto_id(submission_id) and is_ecto_id(question_id) do
-    if role in (@grading_roles ++ @see_all_submissions_roles) do
-      answer_query =
-        Answer
-        |> where(submission_id: ^submission_id)
-        |> where(question_id: ^question_id)
+      when is_ecto_id(submission_id) and is_ecto_id(question_id) and
+             (role in @grading_roles or role in @see_all_submissions_roles) do
+    answer_query =
+      Answer
+      |> where(submission_id: ^submission_id)
+      |> where(question_id: ^question_id)
 
-      answer_query =
-        if role in @grading_roles do
-          students = Cadet.Accounts.Query.students_of(grader)
+    # checks if role is in @grading_roles or @see_all_submissions_roles
+    answer_query =
+      if role in @grading_roles do
+        students = Cadet.Accounts.Query.students_of(grader)
 
-          answer_query
-          |> join(:inner, [a], s in assoc(a, :submission))
-          |> join(:inner, [a, s], t in subquery(students), t.id == s.student_id)
-        else
-          answer_query
-        end
-
-      answer = Repo.one(answer_query)
-
-      with {:answer_found?, true} <- {:answer_found?, is_map(answer)},
-           {:valid, changeset = %Ecto.Changeset{valid?: true}} <-
-             {:valid, Answer.grading_changeset(answer, attrs)},
-           {:ok, _} <- Repo.update(changeset) do
-        {:ok, nil}
+        answer_query
+        |> join(:inner, [a], s in assoc(a, :submission))
+        |> join(:inner, [a, s], t in subquery(students), t.id == s.student_id)
       else
-        {:answer_found?, false} ->
-          {:error, {:bad_request, "Answer not found or user not permitted to grade."}}
-
-        {:valid, changeset} ->
-          {:error, {:bad_request, full_error_messages(changeset.errors)}}
-
-        {:error, _} ->
-          {:error, {:internal_server_error, "Please try again later."}}
+        answer_query
       end
+
+    answer = Repo.one(answer_query)
+
+    with {:answer_found?, true} <- {:answer_found?, is_map(answer)},
+         {:valid, changeset = %Ecto.Changeset{valid?: true}} <-
+           {:valid, Answer.grading_changeset(answer, attrs)},
+         {:ok, _} <- Repo.update(changeset) do
+      {:ok, nil}
     else
-      {:error, {:unauthorized, "User is not permitted to grade."}}
+      {:answer_found?, false} ->
+        {:error, {:bad_request, "Answer not found or user not permitted to grade."}}
+
+      {:valid, changeset} ->
+        {:error, {:bad_request, full_error_messages(changeset.errors)}}
+
+      {:error, _} ->
+        {:error, {:internal_server_error, "Please try again later."}}
     end
+  end
+
+  def update_grading_info(
+        %{submission_id: submission_id, question_id: question_id},
+        _,
+        _
+      )
+      when is_ecto_id(submission_id) and is_ecto_id(question_id) do
+    {:error, {:unauthorized, "User is not permitted to grade."}}
   end
 
   defp find_submission(user = %User{}, assessment = %Assessment{}) do
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -15,8 +15,6 @@ defmodule Cadet.Assessments do
   @xp_early_submission_max_bonus 100
   @xp_bonus_assessment_type ~w(mission sidequest)a
   @submit_answer_roles ~w(student)a
-  @grading_role :staff
-  @see_all_submissions_role :admin
   @see_all_submissions_roles [:staff, :admin]
   @open_all_assessment_roles ~w(staff admin)a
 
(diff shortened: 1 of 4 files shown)


indocomsoft:
@rrtheonlyone do make the changes as discussed on telegram :)

rrtheonlyone:
Yup have worked on the changes!

Do see if its better now

lib/cadet/assessments/assessments.ex
```diff
@@ -448,59 +449,40 @@ defmodule Cadet.Assessments do
...
+    cond do
+      role in @grading_roles ->
+        {:ok, submissions_by_group(grader, submission_query)}
 
+      role in @see_all_submissions_roles ->
         submissions =
-          submission_query
-          |> join(:inner, [s], st in subquery(students), s.student_id == st.id)
-          |> Repo.all()
+          (group_only && role != @admin_role && submissions_by_group(grader, submission_query)) ||
```
indocomsoft: The intent of this line is not too clear. Perhaps this is better?
```elixir
if group_only and role != @admin_role
  submissions_by_group(grader, submission_query)
else
  Repo.all(submission_query)
```

lib/cadet/assessments/assessments.ex
```diff
@@ -422,7 +423,7 @@ defmodule Cadet.Assessments do
 
   @spec all_submissions_by_grader(%User{}) ::
           {:ok, [%Submission{}]} | {:error, {:unauthorized, String.t()}}
-  def all_submissions_by_grader(grader = %User{role: role}) do
+  def all_submissions_by_grader(grader = %User{role: role}, group_only \\ false) do
```
indocomsoft: Can add a `@doc` documenting this function, and explaining `group_only` param

lib/cadet/assessments/assessments.ex
```diff
@@ -448,59 +449,40 @@ defmodule Cadet.Assessments do
...
-        {:ok, answers}
-
-      @see_all_submissions_role ->
-        answers =
-          answer_query
-          |> Repo.all()
-          |> Enum.sort_by(& &1.question.display_order)
-
-        {:ok, answers}
+    if role in @see_all_submissions_roles do
```
indocomsoft: You'll need to restore the old one and use
```elixir
cond do
  role in @grading_roles ->
    ...
  role in @see_all_submission_roles ->
    ...
  true ->
    ...
```

lib/cadet/assessments/assessments.ex
```diff
@@ -517,26 +499,29 @@ defmodule Cadet.Assessments do
...
             |> join(:inner, [a], s in assoc(a, :submission))
             |> join(:inner, [a, s], t in subquery(students), t.id == s.student_id)
+            |> Repo.one()
 
-          @see_all_submissions_role ->
+          role in @see_all_submissions_roles ->
             answer_query
+            |> Repo.one()
+
+          true ->
```
indocomsoft: This is not even possible since we have made sure that `role in (@grading_roles ++ @see_all_submissions_roles)`, so remove this part, then move the `Repo.one()` to outside the cond.

Anyway, this should've been a `nil` as we're doing `Repo.one()` and not `Repo.all()`

lib/cadet_web/controllers/grading_controller.ex
```diff
@@ -4,6 +4,20 @@ defmodule CadetWeb.GradingController do
...
+        render(conn, "index.json", submissions: submissions)
+
+      {:error, {status, message}} ->
+        conn
+        |> put_status(status)
+        |> text(message)
+    end
+  end
+
   def index(conn, _) do
```
indocomsoft: This code is not DRY. Instead what you can do is:
```elixir
def index(conn, _) do
  index(conn, %{"group" => false})
end
```

lib/cadet_web/controllers/grading_controller.ex
```diff
@@ -4,6 +4,20 @@ defmodule CadetWeb.GradingController do
 
   alias Cadet.Assessments
 
+  def index(conn, %{"group" => group}) do
```
indocomsoft: Add a guard clause `when is_boolean(group)`

lib/cadet_web/controllers/grading_controller.ex
```diff
@@ -74,12 +88,16 @@ defmodule CadetWeb.GradingController do
...
 
-    summary("Get a list of all submissions with current user as the grader. ")
+    summary("Get a list of all submissions")
 
     security([%{JWT: []}])
 
     produces("application/json")
 
+    parameters do
+      group(:body, :boolean, "Boolean to show group under grader", required: false)
```
indocomsoft: `"Show only students in the grader's group when true"`

test/cadet_web/controllers/grading_controller_test.exs
```diff
@@ -440,6 +547,43 @@ defmodule CadetWeb.GradingControllerTest do
     end
   end
 
+  describe "GET /group, staff" do
```
indocomsoft: Need to change the description

test/cadet_web/controllers/grading_controller_test.exs
```diff
@@ -440,6 +547,43 @@ defmodule CadetWeb.GradingControllerTest do
...
+    @tag authenticate: :staff
```
indocomsoft: Please do add one for each of the roles (missing `:student` and `:admin`

lib/cadet/assessments/assessments.ex
indocomsoft: Not sure why you need the `role != @admin-role` though? An admin would never be part of a group so he would of course get 0 results. I think that's logical?

lib/cadet/assessments/assessments.ex
```diff
@@ -15,8 +15,9 @@ defmodule Cadet.Assessments do
   @xp_early_submission_max_bonus 100
   @xp_bonus_assessment_type ~w(mission sidequest)a
   @submit_answer_roles ~w(student)a
-  @grading_role :staff
-  @see_all_submissions_role :admin
+  @grading_roles ~w()a
+  @admin_role :admin
```
indocomsoft: Not sure what `@admin_role` is used for

lib/cadet/assessments/assessments.ex
rrtheonlyone: It was because the front-end by default will query `/grading/?group=true`. So, an admin will not see any results until he ticks the checkbox. 

Is there a better way to do that?

lib/cadet/assessments/assessments.ex
rrtheonlyone: Done!

lib/cadet/assessments/assessments.ex
rrtheonlyone: Ahh yup! Missed this out

Done!

lib/cadet/assessments/assessments.ex
rrtheonlyone: Yup was trying to find a clean way of implementing this. Improved!

lib/cadet_web/controllers/grading_controller.ex
rrtheonlyone: Yup done!

lib/cadet_web/controllers/grading_controller.ex
rrtheonlyone: Done

lib/cadet_web/controllers/grading_controller.ex
rrtheonlyone: Done

test/cadet_web/controllers/grading_controller_test.exs
rrtheonlyone: Done

test/cadet_web/controllers/grading_controller_test.exs
rrtheonlyone: Okay! Added 2 more test cases

lib/cadet/assessments/assessments.ex
rrtheonlyone: It was used under the `role != @admin_role` (refer to PR comment above)

I am not sure if it is really needed actually?

lib/cadet/assessments/assessments.ex
indocomsoft: Or instead, you can make the change in frontend instead to request the right thing based on the role.

I'm pretty sure GET /user gives back role

By default show all submission should be checked for admin right?

Cos if not the behaviour would be inconsistent. It's logical that specifying group to be true for any non student not the leader of any group to have 0 submission.

lib/cadet/assessments/assessments.ex
```diff
@@ -15,8 +15,8 @@ defmodule Cadet.Assessments do
   @xp_early_submission_max_bonus 100
   @xp_bonus_assessment_type ~w(mission sidequest)a
   @submit_answer_roles ~w(student)a
-  @grading_role :staff
-  @see_all_submissions_role :admin
+  @grading_roles ~w()a
+  @see_all_submissions_roles [:staff, :admin]
```
indocomsoft: Use the sigil syntax

lib/cadet/assessments/assessments.ex
```diff
@@ -517,26 +525,24 @@ defmodule Cadet.Assessments do
...
-      answer =
-        role
-        |> case do
-          @grading_role ->
-            students = Cadet.Accounts.Query.students_of(grader)
-
-            answer_query
-            |> join(:inner, [a], s in assoc(a, :submission))
-            |> join(:inner, [a, s], t in subquery(students), t.id == s.student_id)
+      answer_query =
```
indocomsoft: Try as much not to do re-assignment in Elixir. What I meant is
```elixir
answer =
  cond do
    role in @grading_roles ->
      ...
    role in @see_all_submissions_role ->
      ...
  end
  |> Repo.one()
```

lib/cadet_web/controllers/grading_controller.ex
```diff
@@ -74,12 +78,21 @@ defmodule CadetWeb.GradingController do
   swagger_path :index do
     get("/grading")
 
-    summary("Get a list of all submissions with current user as the grader. ")
+    summary("Get a list of all submissions")
```
indocomsoft: Not sure why this line needs to be changed.

indocomsoft:
Also, if you can merge from/rebase on `master`

lib/cadet_web/controllers/grading_controller.ex
rrtheonlyone: Yup reverted (wanted to say that by default it gives all submissions but they can infer that from the params)

lib/cadet/assessments/assessments.ex
rrtheonlyone: Changed as per discussion

lib/cadet/assessments/assessments.ex
rrtheonlyone: done!

PR #264, opened 2018-09-06
[WIP] Add plagiarism check

juxd:
Added a plagiarism check script, to be used with [this script](https://github.com/rafaelperes/TA_CS1101S/tree/new-cadet-db)

This is a work in progress. WIll need advice on how to organise the files.


lib/cadet/jobs/autograder/grading_job.ex
```diff
@@ -10,7 +10,7 @@ defmodule Cadet.Autograder.GradingJob do
   require Logger
 
   alias Cadet.Assessments.{Answer, Assessment, Question, Submission}
-  alias Cadet.Autograder.Utilities
+  alias Cadet.Autograder.{Utilities, PlagiarismChecker}
```
indocomsoft: Sort alphabetically

lib/cadet/jobs/autograder/grading_job.ex
```diff
@@ -88,6 +92,14 @@ defmodule Cadet.Autograder.GradingJob do
     end
   end
 
+  def assessment_requires_plagiarism_check(assessment = %Assessment{}) do
+    Question
+    |> where(assessment_id: ^assessment.id)
+    |> select([:type])
```
indocomsoft: I think it's better to have this as `|> select([q], q.type)` and later on `|> Enum.any?(& &1 == :programming)` to save on memory not having to load the whole struct.

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,63 @@
...
+  @plagiarism_script_path "../grader/TA_CS1101S/mosspy_submission.py"
+
+  def perform(assessment_id) when is_ecto_id(assessment_id) do
+    Logger.info("Running plagiarism check on Assessment #{assessment_id}")
+
+    script_result =
+      System.cmd("python", [@plagiarism_script_path, "--assessment_id", to_string(assessment_id)])
+
+    script_result
+    |> elem(1)
```
indocomsoft: Use pattern matching instead on line 21.

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  defp zip_results(assessment_id) do
+    System.cmd("zip", [
+      "-r",
+      "assessment_#{assessment_id}.zip",
+      "submissions/assessment#{assessment_id}/report/",
+      "submissions/assessment#{assessment_id}/assessment_report_#{assessment_id}.html"
+    ])
+
+    assessment_id
```
indocomsoft: Might wanna check whether the `zip` command fails (check the exit value)

lib/cadet/jobs/autograder/grading_job.ex
```diff
@@ -88,6 +92,14 @@ defmodule Cadet.Autograder.GradingJob do
     end
   end
 
+  def assessment_requires_plagiarism_check(assessment = %Assessment{}) do
+    Question
+    |> where(assessment_id: ^assessment.id)
+    |> select([q], q.type)
+    |> Repo.all()
+    |> Enum.any?(&(&1 == :programming))
```
indocomsoft: Actually, better to not even load this into memory. Instead do:
```elixir
Question
|> where(assessment_id: ^assessment.id)
|> where(type: ^:programming)
|> Repo.aggregate(:count, :id)
|> case do
  0 -> false
  _ -> true
end
```

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,70 @@
...
+               |> Keyword.get(:bucket_name)
+
+  @plagiarism_script_path :cadet
+                          |> Application.fetch_env!(:plagiarism_check_vars)
+                          |> Keyword.get(:plagiarism_script_path)
+
+  def perform(assessment_id) when is_ecto_id(assessment_id) do
+    Logger.info("Running plagiarism check on Assessment #{assessment_id}")
+
+    {_, exit_result} =
```
indocomsoft: Refer to https://hexdocs.pm/elixir/System.html#cmd/3. Look at the typespec: actually the return tuple is `{output, exit_code}`

So perhaps what you want is `{result, 0} = System.cmd...`

It's your choice whether to pattern match exit_code with `0` or not. If you do, it's in line with Erlang's let it crash philosophy -- will be captured by sentry. Else you should instead do sth like this:
```elixir
case System.cmd(...) do
  {result, 0} ->
    ...
  _ ->
    # handle error
end
```

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,70 @@
...
+      |> select([a], a.title)
+      |> Repo.one()
+
+    response =
+      @bucket_name
+      |> ExAws.S3.put_object("/reports/assessment-#{assessment_title}", File.read!(file_name))
+      |> ExAws.request!()
+
+    # if response status < 400, the transaction was successful.
+    if Map.get(response, :status_code) < 400 do
```
indocomsoft: What does 400 represent? If it is HTTP, isn't success `200`?

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,70 @@
...
+    else
+      raise inspect(response)
+    end
+  end
+
+  defp zip_results(assessment_id) do
+    {_, zip_result} =
+      System.cmd("zip", [
+        "-r",
+        "assessment_#{assessment_id}.zip",
```
indocomsoft: You should probably put this instead the `submissions` directory so as to not dirty the root directory

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,70 @@
...
+        "submissions/assessment#{assessment_id}/report/",
+        "submissions/assessment#{assessment_id}/assessment_report_#{assessment_id}.html"
+      ])
+
+    if zip_result == 0 do
```
indocomsoft: Pattern match using case instead on the `System.cmd/2`

lib/cadet/jobs/autograder/plagiarism_checker.ex
juxd: yeah, this looks sensible. Will it make sense to output something like a json string as the `result`, and then parse into a `Map`? might want to get multiple value results from the python script.

indocomsoft: So the python script outputs json?

juxd: not yet, but I can make it do so. Will probably be doing that as there are multiple pieces of info to return.

lib/cadet/jobs/autograder/plagiarism_checker.ex
juxd: Do i achieve this with a pipe? I can't pass `mix credo` if i do this haha.
```
 Pipe chain should start with a raw value.
```

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,74 @@
...
+         ]) do
+      {result, 0} ->
+        result
+        |> Jason.decode!()
+        |> Map.get("assessment_id")
+        |> zip_results()
+        |> store()
+
+      _ ->
+        raise "Error running script, please check again"
```
indocomsoft: You might wanna do something like
```elixir
{output, exit_code} ->
  raise "Error running script with exit code #{exit_code}: #{output}"
```

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,74 @@
...
+    end
+  end
+
+  defp store(assessment_id) when is_ecto_id(assessment_id) do
+    file_name = "submissions/assessment_#{assessment_id}.zip"
+
+    assessment_title =
+      Cadet.Assessments.Assessment
```
indocomsoft: Please alias this

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,74 @@
...
+      |> where(id: ^assessment_id)
```
indocomsoft: We prefer `Assessment |> Repo.get(id)` over `Assessment |> where(id: ^id) |> Repo.one()`

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,74 @@
...
+      |> select([a], a.title)
+      |> Repo.one()
+
+    response =
+      @bucket_name
+      |> ExAws.S3.put_object("/reports/assessment-#{assessment_title}.zip", File.read!(file_name))
+      |> ExAws.request!()
+
+    if Map.get(response, :status_code) == 200 do
+      File.rm(file_name)
```
indocomsoft: Isn't this redundant?

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,74 @@
...
+
+  defp zip_results(assessment_id) do
+    case System.cmd("zip", [
+           "-r",
+           "submissions/assessment_#{assessment_id}.zip",
+           "submissions/assessment#{assessment_id}/report/",
+           "submissions/assessment#{assessment_id}/assessment_report_#{assessment_id}.html"
+         ]) do
+      {_, 0} -> assessment_id
+      {_, _} -> raise "Files cannot be zipped. Please check directories."
```
indocomsoft: Same as above, might wanna include output and exit code in the `raise`

indocomsoft:
Also, travis seems to fail because environment is not set. Please resolve this as well

indocomsoft:
Actually @juxd do you have access to push branch to this repo? We use the branch workflow so it'd be better if you can push this as a branch instead of in your own fork.

juxd:
Will transfer this to a branch within this repo.

indocomsoft:
Better make a new PR so I can easily check out without adding another remote

juxd:
moved to #276

PR #276, opened 2018-09-19
Add plagiarism check

juxd:
Migrated from #264 

Sorry for the delay 😅


.gitignore
```diff
@@ -107,4 +107,11 @@ erl_crash.dump
 # VSCode Elixir language support
 .elixir_ls/
 
+# asdf version control
+.tool-versions
```
indocomsoft: Eh? I don't get this leh

config/test.exs
```diff
@@ -36,3 +36,9 @@ config :cadet, Cadet.Auth.Guardian,
   secret_key: "4ZxeVrSvCJlmndrFL7tBpnZsTc/rOQygVIyscAMY1oKKzkKi7hkjXl9F1f28Jap8"
 
 config :arc, definition: Arc.Storage.Local
+
+config :cadet,
```
indocomsoft: Might be better to add this to `config.exs`? Or probably include this as well inside `prod.secrets.exs` as well.

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,77 @@
...
+
+    case System.cmd("python", [
+           @plagiarism_script_path,
+           "--assessment_id",
+           to_string(assessment_id)
+         ]) do
+      {result, 0} ->
+        result
+        |> Jason.decode!()
+        |> Map.get("assessment_id")
```
indocomsoft: I don't quite get why you're doing this. Won't the result from here equal to just calling `to_string(assessment_id)`?

lib/cadet/jobs/autograder/plagiarism_checker.ex
```diff
@@ -0,0 +1,77 @@
...
+
+    response =
+      @bucket_name
+      |> ExAws.S3.put_object("/reports/assessment-#{assessment_title}.zip", File.read!(file_name))
+      |> ExAws.request!()
+
+    if Map.get(response, :status_code) == 200 do
+      File.rm_rf("submissions")
+    else
+      raise inspect(response)
```
indocomsoft: Hmm. If it fails to upload to S3, those zip files will clutter the server then?

indocomsoft: Also you might wanna attempt re-uploading, maybe like 3 times in case connection is flaky?

test/cadet/jobs/autograder/grading_job_test.exs
```diff
@@ -20,6 +20,12 @@ defmodule Cadet.Autograder.GradingJobTest do
     end
   end
 
+  defp assert_plagiarism_for(assessments_list) do
+    for {assessment} <- assessments_list do
```
indocomsoft: The assesments_list is put inside a tuple? Also unless it's nested `Enum.each` or `Enum.map`, I'd prefer using the `Enum` methods since `for` list comprehension can fail silently (when pattern matching failed)

test/cadet/jobs/autograder/plagiarism_checker_test.exs
```diff
@@ -0,0 +1,68 @@
...
+
+        case result do
+          {:ok, _} ->
+            assert result
+                   |> elem(1)
+                   |> MapSet.new()
+                   |> MapSet.equal?(MapSet.new(deleted_files))
+
+          _ ->
+            flunk("Plagiarism Check failed")
```
indocomsoft: Why do you need to flunk tho, just do a pattern match on result and let it crash. It will fail the test. Something like

```elixir
{:ok, deleted_files_actual} = PlagiarismChecker.perform(assessment.id)
assert MapSet.equal?(MapSet.new(deleted_files), MapSet.new(deleted_files_actual))
```

.gitignore
juxd: woops... was supposed to not commit this, then i forgot `.gitignore` itself gets committed ._.

config/test.exs
juxd: hmm, they'll be two different paths... will that make sense?

lib/cadet/jobs/autograder/plagiarism_checker.ex
juxd: The real script itself does the plagiarism checking calls (database connection, writes answers to files, generates a report for said files through an online service). The output of the script is a `json` string that contains the `assessment_id`, but it has a side effect of producing a folder containing all the reports.

indocomsoft: Yea but my question is why do you need to do this? Isn't `result |> Jason.decode!() |> Map.get("assessment_id") == to_string(assessment_id)`? If so why not just use `to_string(assessment_id)`?

I mean proper python script should return non-zero exit code if there's an error, so the pattern matching wouldn't even go here in the first place if it failed right?

config/test.exs
indocomsoft: Well, then the one here would be just for tests, and the one in `prod.secrets.exs` is for production only.

lib/cadet/jobs/autograder/plagiarism_checker.ex
juxd: hmm... seems like ExAws has default retries: https://hexdocs.pm/ex_aws/ExAws.html#module-retries

config/test.exs
indocomsoft: Bump. I think just provide this in 3 separate locations for `test.exs`, `dev.exs`, and `prod.secrets.exs.example` to be provided for production deployment

indocomsoft:
Please merge from/rebase on `master`.

Also, could you check, I'm getting this error during testing:
image

juxd:
Couldn't replicate that test error after rebasing. The request should be going to ExVCR. Any details on which test is throwing this error?

indocomsoft:
@juxd I presume this is stale. Will close after New Year.

juxd:
Ok.

PR #310, opened 2018-12-17
Unsubmit Support (Resolves #271)

### LINKED ISSUE #271: Unsubmit support
There may be legitimate reasons to unsubmit an assignment.

jiachen247:
With reference to #271 

Access Control
- Only avengers can unsubmit submissions by students belonging to their studio group.

Implementation:
- Introduced a new endpoint `unsubmit`
- checks if submission status is submitted and if assessment is still open.
- then proceeds to rollback the status and the auto grading fields.


lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,78 @@ defmodule Cadet.Assessments do
...
 
+  def unsubmit_submission(submission_id, avenger = %User{role: role, id: user_id})
+      when is_ecto_id(submission_id) do
+    if role in @unsubmit_assessment_role do
+      submission =
+        Submission
+        |> where(id: ^submission_id)
+        |> join(:inner, [s], a in assoc(s, :assessment))
+        |> preload([_, a], assessment: a)
+        |> Repo.one()
```
indocomsoft: I think it's better to write it this way instead:
```elixir
        Submission
        |> join(:inner, [s], a in assoc(s, :assessment))
        |> preload([_, a], assessment: a)
        |> Repo.get(submission_id)
```

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,78 @@ defmodule Cadet.Assessments do
...
+
+      with {:submission_found?, true} <- {:submission_found?, is_map(submission)},
+           {:is_open?, true} <- is_open?(submission.assessment),
+           {:status, :submitted} <- {:status, submission.status},
+           {:avenger_of?, true} <- {:avenger_of?, is_avenger_of(avenger, submission.student_id)} do
+        rollback_submission_status(submission)
+        rollback_grading_answers(submission)
+
+        {:ok, nil}
```
indocomsoft: Lmao why not just `:ok`

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,78 @@ defmodule Cadet.Assessments do
...
+        {:status, :attempting} ->
+          {:error, {:bad_request, "Some questions have not been attempted"}}
+
+        {:status, :attempted} ->
+          {:error, {:forbidden, "Assessment has not been submitted"}}
+
+        {:avenger_of, false} ->
+          {:error, {:forbidden, "Only avenger of student is allowed to unsubmit"}}
+
+        _ ->
```
indocomsoft: Just let it crash. I don't think this clause is necessary

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,78 @@ defmodule Cadet.Assessments do
...
+          {:error, {:internal_server_error, "Please try again later."}}
+      end
+    else
+      {:error, {:forbidden, "User is not permitted to answer questions"}}
```
indocomsoft: Did you mean: not permitted to unsubmit?

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,78 @@ defmodule Cadet.Assessments do
...
+    end
+  end
+
+  def is_avenger_of(avenger, student_id) do
```
indocomsoft: I think this should be moved to query.ex instead

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,78 @@ defmodule Cadet.Assessments do
...
+    Cadet.Accounts.Query.students_of(avenger)
+    |> where(id: ^student_id)
```
indocomsoft: I think this looks better:
```elixir
Cadet.Accounts.Query.students_of(avenger)
|> Repo.get(student_id)
|> case do
  nil -> false
  _ -> true
end
```

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,78 @@ defmodule Cadet.Assessments do
...
+  def rollback_submission_status(submission = %Submission{}) do
+    submission
+    |> Submission.changeset(%{status: :attempted, xp_bonus: 0})
+    |> Repo.update()
+  end
+
+  def rollback_grading_answers(submission = %Submission{}) do
+    Answer
+    |> where(submission_id: ^submission.id)
+    |> Repo.update_all(
```
indocomsoft: > `update_all` will not update autogenerated fields like the updated_at columns.

I think it's better to just map the answers and run `update` on them, wrapped in a `transaction`.

lib/cadet/assessments/assessments.ex
```diff
@@ -421,13 +494,13 @@ defmodule Cadet.Assessments do
   end
 
   @doc """
-  Function returning submissions under a grader.
```
indocomsoft: You should check what's causing this. Is it CRLF vs LF? You shld revert these useless changes cos it's noise on the diff haha

lib/cadet_web/controllers/assessments_controller.ex
```diff
@@ -17,6 +17,18 @@ defmodule CadetWeb.AssessmentsController do
     end
   end
 
+  def unsubmit(conn, %{"submissionid" => submission_id}) when is_ecto_id(submission_id) do
```
indocomsoft: Not sure if this should be part of assessments or grading. @tuesmiddt?

lib/cadet_web/controllers/assessments_controller.ex
```diff
@@ -48,6 +60,21 @@ defmodule CadetWeb.AssessmentsController do
     response(404, "Submission not found")
   end
 
+  swagger_path :unsubmit do
+    post("/assessments/{submissionId}/unsubmit")
+    summary("Unsubmit submission for an assessment")
```
indocomsoft: Should specify that this can only be done by avengers of the student.

lib/cadet_web/controllers/assessments_controller.ex
indocomsoft: I would argue that it should be part of grading, because anything that takes in submission_id so far are in grading.

indocomsoft: Also, please add another `unsubmit/2` for when the user does not provide `submissionid`

mix.lock
```diff
@@ -15,7 +15,7 @@
   "crontab": {:hex, :crontab, "1.1.4", "eae8de759d60dec2352ca577890f4cac26bcb907088ffef958f42ebb9c4f4cff", [:mix], [{:ecto, "~> 1.0 or ~> 2.0 or ~> 2.1", [hex: :ecto, repo: "hexpm", optional: true]}], "hexpm"},
   "csv": {:hex, :csv, "2.0.0", "c66fea89ba7862b94901baf0871285e9b73cad89c5fdb57a6386d2adcf29593e", [:mix], [{:parallel_stream, "~> 1.0.4", [hex: :parallel_stream, repo: "hexpm", optional: false]}], "hexpm"},
   "db_connection": {:hex, :db_connection, "1.1.3", "89b30ca1ef0a3b469b1c779579590688561d586694a3ce8792985d4d7e575a61", [:mix], [{:connection, "~> 1.0.2", [hex: :connection, repo: "hexpm", optional: false]}, {:poolboy, "~> 1.5", [hex: :poolboy, repo: "hexpm", optional: true]}, {:sbroker, "~> 1.0", [hex: :sbroker, repo: "hexpm", optional: true]}], "hexpm"},
-  "decimal": {:hex, :decimal, "1.6.0", "bfd84d90ff966e1f5d4370bdd3943432d8f65f07d3bab48001aebd7030590dcc", [], [], "hexpm"},
+  "decimal": {:hex, :decimal, "1.6.0", "bfd84d90ff966e1f5d4370bdd3943432d8f65f07d3bab48001aebd7030590dcc", [:mix], [], "hexpm"},
```
indocomsoft: Heh, this is a funny change. Prolly just checkout the original file?

lib/cadet/assessments/assessments.ex
indocomsoft: Oh, might want to alias `Cadet.Accounts.Query` as well

lib/cadet_web/controllers/assessments_controller.ex
jiachen247: hmmm i put it in assessment  because it seems to undo `finalize_submission`. but point taken we can put it in grading instead.

jiachen247:
dup #312

PR #312, opened 2018-12-25
Unsubmit Support (Resolves #271)

### LINKED ISSUE #271: Unsubmit support

jiachen247:
With reference to #271

Access Control

- Only avengers can unsubmit submissions by students belonging to their studio group.

Implementation:

- Introduced a new endpoint unsubmit
- Checks if submission status is submitted and if assessment is still open.
- then proceeds to rollback the status and the auto grading fields.


lib/cadet/accounts/query.ex
```diff
@@ -26,6 +27,15 @@ defmodule Cadet.Accounts.Query do
     |> where([_, g], g.leader_id == ^id)
   end
 
+  def is_avenger_of(avenger, student_id) do
```
jiachen247: this technically isnt a query since its being executed. is there a better place to put this?

indocomsoft: Ah ok, I guess it's fine then.

indocomsoft: Also, do read the style guide at https://github.com/source-academy/cadet#style-guide. By convention, a function that returns boolean should not have `is_` in front, but rather end with question mark. so `def avenger_of?(avenger, student_id)` is better.

lib/cadet/assessments/answer.ex
```diff
@@ -40,6 +40,14 @@ defmodule Cadet.Assessments.Answer do
     |> validate_xp_grade_adjustment_total()
   end
 
+  def reset(answer, params) do
```
indocomsoft: Since this returns a changeset, you might want to call this `reset_changeset` instead

indocomsoft: You should do `validate_required` as well. I think it's better to refactor the common part of `changeset/2` and `reset_changeset/2` into a private function instead.

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,73 @@ defmodule Cadet.Assessments do
...
+        |> join(:inner, [s], a in assoc(s, :assessment))
+        |> preload([_, a], assessment: a)
+        |> Repo.get(submission_id)
+
+      with {:submission_found?, true} <- {:submission_found?, is_map(submission)},
+           {:is_open?, true} <- is_open?(submission.assessment),
+           {:status, :submitted} <- {:status, submission.status},
+           {:avenger_of?, true} <-
+             {:avenger_of?, Cadet.Accounts.Query.is_avenger_of(avenger, submission.student_id)} do
+        Ecto.Multi.new()
```
indocomsoft: Be consistent, if you've aliased `Ecto.Multi` then this should be just `Multi`

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,73 @@ defmodule Cadet.Assessments do
...
+            submission
+            |> Submission.changeset(%{status: :attempted, xp_bonus: 0})
+            |> Repo.update()
+          end
+        )
+        |> Multi.run(:rollback_answers, fn _ ->
+          Answer
+          |> where(submission_id: ^submission.id)
+          |> Repo.all()
+          |> Enum.each(fn answer ->
```
indocomsoft: Instead I think you should instead do `Enum.reduce` on the result of `Repo.update/1` and then make sure the function passed to `Multi.run/3` returns the appropriate `{:ok, nil}` or `{:error, ...}` because right now your multi will not terminate and rollback on update error.

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,73 @@ defmodule Cadet.Assessments do
...
+        |> preload([_, a], assessment: a)
+        |> Repo.get(submission_id)
+
+      with {:submission_found?, true} <- {:submission_found?, is_map(submission)},
+           {:is_open?, true} <- is_open?(submission.assessment),
+           {:status, :submitted} <- {:status, submission.status},
+           {:avenger_of?, true} <-
+             {:avenger_of?, Cadet.Accounts.Query.is_avenger_of(avenger, submission.student_id)} do
+        Ecto.Multi.new()
+        |> Multi.run(
```
indocomsoft: Why not use `Multi.update/4`?

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,73 @@ defmodule Cadet.Assessments do
...
+        {:is_open?, false} ->
+          {:error, {:forbidden, "Assessment not open"}}
+
+        {:status, :attempting} ->
+          {:error, {:bad_request, "Some questions have not been attempted"}}
+
+        {:status, :attempted} ->
+          {:error, {:forbidden, "Assessment has not been submitted"}}
+
+        _ ->
```
indocomsoft: Let it crash! I don't think there's a need for this clause. Also you're missing the clause for `{:avenger_of?, false}`

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,73 @@ defmodule Cadet.Assessments do
...
+              comment: nil
+            })
+            |> Repo.update()
+          end)
+
+          {:ok, nil}
+        end)
+        |> Repo.transaction()
+
+        {:ok, nil}
```
indocomsoft: Why not just `:ok`?

lib/cadet_web/controllers/grading_controller.ex
```diff
@@ -99,6 +117,21 @@ defmodule CadetWeb.GradingController do
     response(401, "Unauthorised")
   end
 
+  swagger_path :unsubmit do
+    post("/grading/{submissionId}/unsubmit")
+    summary("Unsubmit submission for an assessment by Avenger of student.")
```
indocomsoft: Make it clearer: `Unsubmit submission. Can only be done by the Avenger of a student.`

lib/cadet_web/controllers/grading_controller.ex
```diff
@@ -77,6 +77,24 @@ defmodule CadetWeb.GradingController do
     |> text("Missing parameter")
   end
 
+  def unsubmit(conn, %{"submissionid" => submission_id}) when is_ecto_id(submission_id) do
+    case Assessments.unsubmit_submission(submission_id, conn.assigns.current_user) do
```
indocomsoft: Better to abstract away the `conn.assigns[:current_user]` like in line 50.

lib/cadet/assessments/assessments.ex
```diff
@@ -371,6 +372,73 @@ defmodule Cadet.Assessments do
...
+          {:error, {:not_found, "Submission not found"}}
+
+        {:is_open?, false} ->
+          {:error, {:forbidden, "Assessment not open"}}
+
+        {:status, :attempting} ->
+          {:error, {:bad_request, "Some questions have not been attempted"}}
+
+        {:status, :attempted} ->
+          {:error, {:forbidden, "Assessment has not been submitted"}}
```
indocomsoft: I think this should be `:bad_request` instead

lib/cadet/assessments/assessments.ex
jiachen247: I was trying to be consistent with the rest of the functions like `finalise_submission` that seem to return {:ok, nil}

indocomsoft: Oh. Hmm ok it's fine then.

lib/cadet/assessments/answer.ex
jiachen247: `validate_required` seems to need to preload the question in. Can we just assume since data has been added that it has already passed the validation initially? (we are not introducing new data) thats the reason why `reset_changeset` was created instead of just using changeset.

indocomsoft: I don't think that's a good idea. Why not just use the original `changeset/2`?

indocomsoft:
Btw you should merge from master or rebase on master :)

jiachen247:
okayss!! Will write test cases and rebase and i think we can try to merge this tmr if all goes well.

martin-henz:
Is this PR ready to be merged?

geshuming:
LGTM works fine on my end

PR #358, opened 2019-03-28
mix deps.update --all

jiachen247:


mix.exs
```diff
@@ -73,7 +73,7 @@ defmodule Cadet.Mixfile do
       # TODO: Remove the override once ex_aws released the new version
       #       without the dependency on xml_builder. Waste my time urgh
       {:xml_builder, "~> 2.0", override: true},
-      {:credo, "~> 1.0", only: [:dev, :test], runtime: false},
+      {:credo, "== 1.0.3", only: [:dev, :test], runtime: false},
```
indocomsoft: You can change this to be semver `~>` again, because now `mix.lock` has recorded the `1.0.3` version. Dependabot will be a bit noisy but you can just ignore it and it will inform you of `1.0.5`

jiachen247: okay lets do this!

PR #364, opened 2019-04-17
Luminus Authentication (Resolves  #337)

### LINKED ISSUE #337: IVLE is being deprecated next semester
Our only log in solution is using ivle right now.

There are 2 possible solutions that I see:
1. Add a new way of logging in using username and password. This will require a non-trivial amount of effort
2. I have created a library in elixir to connect to luminus which could perhaps be used in place of the IVLE authorization method. However, it is based on reverse engineering their angular frontend -- they have been postponing the release of their public API since March 2018. There is a non-zero probability that this API will change some time in the future. The library is located at https://github.com/indocomsoft/fluminus

jiachen247:
This is a PR to port authentication from IVLE to LumiNUS.
closes #337 

As discussed with Prof Martin, to remove announcements from cadet.


jiachen247:
Accompanying PR on the frontend can be found here!
https://github.com/source-academy/cadet-frontend/pull/534

lib/cadet/accounts/luminus.ex
```diff
@@ -0,0 +1,211 @@
...
+        {:error, :bad_request}
+    end
+  end
+
+  defp parse_modules(modules) do
+    cs1101s =
+      modules["data"]
+      |> Enum.find(fn module ->
+        module["name"] == "CS1101S" and
+          module["id"] != "00000000-0000-0000-0000-000000000000"
```
indocomsoft: Is this still needed tho? AFAIK they also send the semester of module, so you can check that instead.

lib/cadet/accounts/luminus.ex
```diff
@@ -0,0 +1,211 @@
...
+    case api_call("module", Token: token) do
+      {:ok, modules} ->
+        parse_modules(modules)
+
+      {:error, _} ->
+        {:error, :bad_request}
+    end
+  end
+
+  defp parse_cs1101s(cs1101s) do
```
indocomsoft:
```suggestion
  defp parse_cs1101s({:ok, %{"access" => %{"access_Full" => true}}), do: {:ok, :admin}
  defp parse_cs1101s({:ok, %{"access" => %{"access_Create" => true}}), do: {:ok, :staff}
  defp parse_cs1101s({:ok, %{"access" => %{"access_Read" => true}}), do: {:ok, :student}
  defp parse_cs1101s(_), do: {:error, :bad_request}
```

indocomsoft: And delete the rest lol

indocomsoft: Also why is this a separate function? Why not put this inside `parse_modules/1`? I think that would still be fine.

lib/cadet/accounts/luminus.ex
```diff
@@ -0,0 +1,211 @@
...
+        grant_type: "authorization_code",
+        code: code,
+        redirect_uri: @redirect_url
+      }
+      |> URI.encode_query()
+
+    headers = [{"Content-type", "application/x-www-form-urlencoded"}]
+
+    case HTTPoison.post(@api_token_url, queries, headers) do
+      {:ok, %{body: body, status_code: 200}} when body != ~s("") ->
```
indocomsoft: Why do you need the guard clause `when body != ~s("")`?

Just let it crash!

test/cadet/accounts/accounts_test.exs
```diff
@@ -134,73 +133,46 @@ defmodule Cadet.AccountsTest do
...
-      end
-    end
-
-    test ~s(user has unknown permission "A") do
-      use_cassette "accounts/sign_in#9", custom: true do
-        assert {:error, :bad_request} = Accounts.sign_in("e012345", @token)
-      end
-    end
-
     test ~s(user cs1101s module is inactive) do
```
indocomsoft: U sure that inactive mods has that code? Maybe check with LumiNUS? Or like I said earlier check the module's semester instead

test/cadet/accounts/accounts_test.exs
```diff
@@ -134,73 +133,46 @@ defmodule Cadet.AccountsTest do
...
+      use_cassette "accounts/sign_in#6", custom: true do
+        assert {:error, :bad_request} = Accounts.sign_in("e012345", "TOM", @token)
       end
     end
 
     test ~s(user does not read cs1101s) do
-      use_cassette "accounts/sign_in#11", custom: true do
-        assert {:error, :bad_request} = Accounts.sign_in("e012345", @token)
+      use_cassette "accounts/sign_in#7", custom: true do
+        assert {:error, :bad_request} = Accounts.sign_in("e012345", "TOM", @token)
```
indocomsoft: Hmm now that I think about it, would 403 `:forbidden` be a better code?

test/cadet/accounts/luminus_test.exs
```diff
@@ -0,0 +1,106 @@
...
+  @token if System.get_env("CODE"),
+           do: elem(Luminus.fetch_luminus_token(System.get_env("CODE")), 1),
+           else: "token"
+
+  setup_all do
+    HTTPoison.start()
+  end
+
+  describe "Do an API call; methods with empty string for invalid token" do
+    test "With one parameter token" do
```
indocomsoft: one parameter token?

Also why are the tests all calling different endpoints?

test/cadet/accounts/luminus_test.exs
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    test "With an invalid token" do
+      use_cassette "luminus/api_call#3" do
+        assert {:error, :bad_request} = Luminus.api_call("UserName_Get", Token: @token <> "Z")
+      end
+    end
+  end
+
+  describe "Fetch an details" do
```
indocomsoft:
```suggestion
  describe "Fetch details" do
```

test/cadet/accounts/luminus_test.exs
```diff
@@ -0,0 +1,106 @@
...
+      use_cassette "luminus/fetch_details#1" do
+        assert {:ok, nusnet_id, name} = Luminus.fetch_details(@token)
+        assert String.length(nusnet_id) > 0
+        assert String.length(name) > 0
+      end
+    end
+
+    test "Using an invalid token" do
+      use_cassette "luminus/fetch_details#2" do
+        assert {:error, resp} = Luminus.fetch_details(@token <> "Z")
```
indocomsoft:
```suggestion
        assert {:error, :bad_request} = Luminus.fetch_details(@token <> "Z")
```

test/cadet/accounts/luminus_test.exs
```diff
@@ -0,0 +1,106 @@
...
+        assert resp == :bad_request
+      end
+    end
+  end
+
+  describe "Fetch a role" do
+    test "Using a valid token" do
+      use_cassette "luminus/fetch_role#1" do
+        assert {:ok, role} = Luminus.fetch_role(@token)
+        assert Enum.member?([:student, :staff, :admin], role)
```
indocomsoft:
```suggestion
        assert role in [:student, :staff, :admin]
```

test/cadet/accounts/luminus_test.exs
```diff
@@ -0,0 +1,106 @@
...
+
+  setup_all do
+    HTTPoison.start()
+  end
+
+  describe "Do an API call; methods with empty string for invalid token" do
+    test "With one parameter token" do
+      use_cassette "luminus/api_call#1" do
+        assert {:ok, resp} = Luminus.api_call("user/Profile", Token: @token)
+        assert resp != %{}
```
indocomsoft: Not sure what you're trying to do here

test/cadet/accounts/luminus_test.exs
```diff
@@ -0,0 +1,106 @@
...
+      end
+    end
+
+    test ~s(Create Access maps to :staff) do
+      use_cassette "luminus/fetch_role#4", custom: true do
+        assert {:ok, :staff} = Luminus.fetch_role(@token)
+      end
+    end
+
+    test ~s(Permission "S" maps to :student) do
```
indocomsoft: We don't have permission "S" anymore right?

test/cadet_web/controllers/auth_controller_test.exs
```diff
@@ -1,21 +1,20 @@
...
-  `Cadet.Accounts.IVLE`) and environment variable TOKEN (used here). Don't
+  to set the config variables `luminus_api_key`, `luminus_client_id`,
+  `luminus_client_secret` and `luminus_redirect_url` (used as a module attribute
+  in `Cadet.Accounts.Luminus`) and environment variable CODE (used here). Don't
   forget to delete the cassette files, otherwise ExVCR will not override the
-  cassettes. You can set the TOKEN environment variable like so,
+  cassettes. You can set the CODE environment variable like so,
 
-    TOKEN=very_long_token_here mix test
+    CODE=auth_code_goes_here
```
indocomsoft:
```suggestion
    CODE=auth_code_goes_here mix test
```

lib/cadet/accounts/luminus.ex
indocomsoft: I'm sure you know how to figure it out: LumiNUS provides an API to get what is the current term: https://github.com/indocomsoft/fluminus/blob/master/lib/fluminus/api.ex#L57 (you can cache this somewhere)

So then, you can check whether the CS1101S module is current term: https://github.com/indocomsoft/fluminus/blob/master/lib/fluminus/api.ex#L141

test/cadet/accounts/accounts_test.exs
```diff
@@ -1,30 +1,30 @@
...
 
   use Cadet.DataCase
   use ExVCR.Mock, adapter: ExVCR.Adapter.Hackney
 
   alias Cadet.{Accounts, Repo}
-  alias Cadet.Accounts.{Query, User}
+  alias Cadet.Accounts.{Query, Luminus, User}
 
-  @token if System.get_env("TOKEN"), do: System.get_env("TOKEN"), else: "token"
+  @code System.get_env("CODE") || nil
```
indocomsoft:
```suggestion
  @code System.get_env("CODE")
```
`System.get_env/1` would already return `nil` if the environment variable is undefined

test/cadet/accounts/accounts_test.exs
```diff
@@ -134,73 +134,46 @@ defmodule Cadet.AccountsTest do
...
       use_cassette "accounts/sign_in#2" do
-        assert {:error, :bad_request} == Accounts.sign_in("A0123456", "t0k3n")
+        assert {:error, :bad_request} == Accounts.sign_in("A0123456", "TOM", "t0k3n")
       end
     end
 
     test "invalid nusnet id" do
       use_cassette "accounts/sign_in#3" do
-        assert {:error, :internal_server_error} == Accounts.sign_in("", @token)
+        assert {:error, :internal_server_error} == Accounts.sign_in("", "TOM", @token)
```
indocomsoft: By right, invalid NUSNET id should return `:bad_request` right?

test/cadet_web/controllers/auth_controller_test.exs
```diff
@@ -26,7 +25,7 @@ defmodule CadetWeb.AuthControllerTest do
   alias Cadet.Auth.Guardian
   alias CadetWeb.AuthController
 
-  @token if System.get_env("TOKEN"), do: System.get_env("TOKEN"), else: "token"
+  @code System.get_env("CODE") || "code"
```
indocomsoft: Do you really need the `|| "code"`? What if you just remove that and let `@code` be `nil`

jiachen247:
```
Object.defineProperty(Task.prototype, "isPast", {
        get: function () {
            return __WEBPACK_IMPORTED_MODULE_3_moment__().isAfter(this.endDate);
        },
        enumerable: true,
        configurable: true
    });
```
taken from https://luminus.nus.edu.sg/main.d1c031e4d375a32b95d6.bundle.js

Looks like the luminus web client parses and checks the end date to see if the module is over. looks like we have to do the same!

indocomsoft:
Sure. For `fluminus` I checked against the current term, but that's 1 additional API call that luminus backend might not be able to afford. Let's do that then (check end date).

jiachen247:
@indocomsoft its actually in the same api call!! the end date is given so its pretty trivial to parse it with timex. this is how the luminus web client does it just wanna make sure we're not making any assumptions about the api haha

jiachen247:
```
Object.defineProperty(CourseModule.prototype, "isStudent", {
        get: function () {
            return this.access && (this.access.accessRead &&
                !this.access.accessCreate &&
                !this.access.accessDelete &&
                !this.access.accessFull &&
                !this.access.accessSettingsRead &&
                !this.access.accessSettingsUpdate &&
                !this.access.accessUpdate);
        },
        enumerable: true,
        configurable: true
    });
    Object.defineProperty(CourseModule.prototype, "isReadManager", {
        get: function () {
            return this.access && (this.access.accessRead &&
                !this.access.accessCreate &&
                !this.access.accessDelete &&
                !this.access.accessFull &&
                this.access.accessSettingsRead &&
                !this.access.accessSettingsUpdate &&
                !this.access.accessUpdate);
        },
        enumerable: true,
        configurable: true
    });
    Object.defineProperty(CourseModule.prototype, "isManager", {
        get: function () {
            return this.access && (this.access.accessRead &&
                this.access.accessCreate &&
                this.access.accessDelete &&
                !this.access.accessFull &&
                this.access.accessSettingsRead &&
                this.access.accessSettingsUpdate &&
                this.access.accessUpdate);
        },
        enumerable: true,
        configurable: true
    });
    Object.defineProperty(CourseModule.prototype, "isOwner", {
        get: function () {
            return this.access && (this.access.accessRead &&
                this.access.accessCreate &&
                this.access.accessDelete &&
                this.access.accessFull &&
                this.access.accessSettingsRead &&
                this.access.accessSettingsUpdate &&
                this.access.accessUpdate);
        },
        enumerable: true,
        configurable: true
    });
```

And this is how roles and permissions are assigned and handled under the hood on the luminus web client.

@martin-henz how do you suggest we go around mapping the permissions?

I suggest we do

Student -> student
ReadManager -> staff (we can use this for the TAs)
Owner & Manager -> admin role

test/cadet/accounts/accounts_test.exs
jiachen247: Hmmm I think it should be a 500 internal server error leh. Since nusnet id comes directly from luminus it should never be invalid or wrong. If it is either our server or their server has an inconsistent state. haha but im open to throw a bad request too.

lib/cadet/accounts/accounts.ex
```diff
@@ -118,25 +118,27 @@ defmodule Cadet.Accounts do
...
       # user is not registered in our database
-      with {:ok, name} <- IVLE.fetch_name(token),
-           {:ok, role} <- IVLE.fetch_role(token),
+      with {:ok, role} <- Luminus.fetch_role(token),
            {:ok, _} <- register(%{name: name, nusnet_id: nusnet_id}, role) do
-        sign_in(nusnet_id, token)
+        sign_in(nusnet_id, name, token)
       else
+        {:error, :forbidden} ->
+          # Luminus.fetch_*/1 responds with :forbidden if student does not read CS1101S
```
indocomsoft:
```suggestion
          # Luminus.fetch_role/1 responds with :forbidden if student does not read CS1101S
```

lib/cadet/accounts/accounts.ex
```diff
@@ -118,25 +118,27 @@ defmodule Cadet.Accounts do
...
+          {:error, :forbidden}
+
         {:error, :bad_request} ->
-          # IVLE.fetch_*/1 responds with :bad_request if token is invalid
+          # Luminus.fetch_*/1 responds with :bad_request if token is invalid
```
indocomsoft:
```suggestion
          # Luminus.fetch_role/1 responds with :bad_request if token is invalid
```

lib/cadet/accounts/luminus.ex
```diff
@@ -47,11 +74,11 @@ defmodule Cadet.Accounts.Luminus do
       {:ok, "12742174091894830298409823098098"}
   """
 
-  def fetch_luminus_token!(nil) do
+  def fetch_luminus_token_or_return_default(nil) do
```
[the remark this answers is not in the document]
indocomsoft: Usually the way someone does it is:

```elixir
def something!(args) do
  {:ok, result} = something(args)
  result
end
```

Because this will raise `MatchError` if result is bad.

lib/cadet/accounts/luminus.ex
```diff
@@ -0,0 +1,234 @@
...
+      {:error, _} ->
+        {:error, :bad_request}
+    end
+  end
+
+  defp moduleActive?(endDate) do
+    Timex.before?(Timex.now(), Timex.parse!(endDate, "{ISO:Extended}"))
+  end
+
+  defp cs1101s?(name) do
```
indocomsoft: Why does this need to be a function? I think you should inline this. It's clearer that way.

test/cadet_web/controllers/auth_controller_test.exs
indocomsoft: Not sure why you need the `|| "code"` still.

config/config.exs
```diff
@@ -78,7 +78,7 @@ config :guardian, Guardian.DB,
   # default: 60 minute
   sweep_interval: 60
 
-# Import secrets, such as the IVLE key, or guest account credentials
+# Import secrets, such as the LumiNUS key, or guest account credentials
 # The secret.exs file holds secrets that are useful even in development, and
 # so is kept separate from the prod.secret.exs file, which holds secrets useful
 # only for configuring the production build.
```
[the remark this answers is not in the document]
jiachen247: why are we removing this!! haha arent we still polling the private bitbucket at this interval if the updater flag is turned on?

lib/cadet/accounts/luminus.ex
jiachen247: its about time we externalize this into the config file so we can use XS1101S for staging and dev purposes! haha another PR!

test/cadet_web/controllers/auth_controller_test.exs
jiachen247: we need it as a placeholder if code is not given during the test so the tests will still run!

jiachen247:
After talking to @indocomsoft we agreed following map.get/3 with a default value would be the cleanest way to do Luminus.fetch_luminus_token/2.

https://hexdocs.pm/elixir/Map.html#get/3

config/config.exs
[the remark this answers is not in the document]
jiachen247: just found this!! Its safe to remove.
`{"@hourly", {Mix.Tasks.Cadet.Assessments.Update, :run, [nil]}},`

PR #381, opened 2019-05-30
Add autograding results to gradingview

geshuming:


lib/cadet_web/views/assessments_view.ex
```diff
@@ -141,7 +141,7 @@ defmodule CadetWeb.AssessmentsView do
     end
   end
 
-  defp build_result(result) do
+  def build_result(result) do
```
jiachen247: I think this belongs in view_helpers now since its being shared?

lib/cadet_web/views/grading_view.ex
```diff
@@ -35,17 +35,26 @@ defmodule CadetWeb.GradingView do
...
-        ),
+      question: &build_grading_question/1,
       solution: &(&1.question.question["solution"] || ""),
       grade: &build_grade/1
     })
   end
 
+  defp build_grading_question(answer) do
+    results =
+      case answer.autograding_results do
```
jiachen247: i think using pattern matching here would be more expressive!! something like

```elixir
defp build_autograding_results(nil), do: nill
defp build_autograding_results(results) do
    Enum.map(results, &CadetWeb.AssessmentsView.build_result/1)
end
```

then just call the function below or put it in a variable first also can!!

lib/cadet_web/views/assessments_view.ex
geshuming: It should still be fine leaving it under assessments_view? At most only grading_view and assessments_view require build_result so I don't think it makes sense to isolate the function.

lib/cadet_web/views/grading_view.ex
geshuming: Refactored as requested.

lib/cadet_web/views/assessments_view.ex
jiachen247: the pattern we have been adopting is all shared methods go into the view helpers so our views dont couple or depend on other views!! lets try to be consistent with the codebase okay!!

Issue #383, opened 2019-06-05
CS1101S is hardcoded into the auth check


jiachen247:
The `CS1101S` module code is currently hard coded in the ivle/luminus checks.

This tightly couples cadet to cs1101s. In addition, prof @martin-henz  has assigned us a staging module for us to play around with. I propose moving the module code into general config to make the backend more extensible.

to add this to deployment handbook as well.

martin-henz:
Yes, this is crucial, not just for testing: I'll be teaching a summer workshop module called 
SWS3012 Computational Pearls with Functional Programming
where we cover the functional programming subset of CS1101S. We will (of course) use the Source Academy, and the module is already created in LumiNUS.

jiachen247:
Closed with #384

PR #384, opened 2019-06-05
Externalize module code to config file

jiachen247:
To allow for custom modules codes to be used with luminus authentication.
checked out from luminus-auth branch


jiachen247:
#383

PR #396, opened 2019-06-21
Practical Assessment Setup (Resolves #361) 

### LINKED ISSUE #361: DevOps: Assessment and Story Infrastructure
We should allow missions to be protected with a URL parameter. We will use the same Source Academy as always, without any need of starting a separate server. We will hide the URL parameter in a shortcut on the desktop of the lab PCs. 

So for example we will have a shortcut "PE" on the desktop, which has the following link:
https://sourceacademy.nus.edu.sg?key=9d8f7g6h5j4k3l2

In this example, the key 9d8f7g6h5j4k3l2 unlocks the PA mission in the assessment section of the Source Academy. The students will need to log in, as usual, using LumiNUS, and go to the PA.

This way, we can:
* avoid installing any other server
* test the PA setup in advance by supplying the correct key, without any time pressure

In addition: log the access of such 'protected' assignment/mission:
- student who accessed and time of access.

jiachen247:
Resolves #361


jiachen247:
@geshuming i disagree leh I think that its important and clearer that the assessment overview knows its a password protected assessment and renders it differently from the normal ones. This can only be done if it knows its password protected throught he passwordProtected flag!

geshuming:
@jiachen247 how about defining a password protected assessment as a 'practical' assessment type, since password protection will be used for practical assessment only, we can use the mission typing to create a different render.

jiachen247:
@martin-henz what do you think? Im open to both I just have been operating under the assumption that the PE will be a special type of mission. But lossely associating them this way we can always leverage the password for other missions that are not PEs for whatever reason. Or we could do both!

martin-henz:
> @martin-henz what do you think? Im open to both I just have been operating under the assumption that the PE will be a special type of mission. But loosely associating them this way we can always leverage the password for other missions that are not PEs for whatever reason. Or we could do both!

I think we should treat these as two separate orthogonal dimensions:
* Assessment type: missions, quests, practicals, paths and contests
* Access: currently "open" or "passwordProtected"

By keeping these access orthogonal, we can easily protect other components with a password. 

So in this spirit, I'd like to argue for an attribute of TASK called `access`, with the following options:
* `passwordProtected`: A password is required by the server, to obtain the assessment
* `open`: Any student can access without password

The `TASK` element would then have a sub-element that literally provides the password:
```
<TASK ....>
   <PASSWORD>mylongpasswordgoeshere</PASSWORD>
</TASK>
```

martin-henz:
@jiachen247 can you update the Assessment XML specs in the wiki of the assessments repo?

jiachen247:
@martin-henz Hi Prof thanks for youre input I totally agree this seems like a cleaner schema with sensible defaults.

I would propose using `public` and `private` as access modifiers as opposed to `open` and `passwordProtected`. The schema will look something like this.

```
<TASK access="private" ....>
   <PASSWORD>mylongpasswordgoeshere</PASSWORD>
</TASK>
```

And to reiterate if access is set to private and no password field is given an empty string will be used for the password. If access is public and password is given password is disregarded. If access is not provided defaults to public.

Note that this access modifier will be an assessment level feature meaning it will span all the assessments types not just practical. We will decide how to handle this on the frontend on a later date!

jiachen247:
Hi @all,

Made all the changes as discussed above. Added a new assessment type `practical`. The code can be cleaned up abit more!! But core functionality are in already. @geshuming Can you help to take a look through and see if there are any red flags!! thanks!!

We can start working on the frontend already!!

lib/cadet/jobs/xml_parser.ex
```diff
@@ -169,6 +184,14 @@ defmodule Cadet.Updater.XMLParser do
       :error
   end
 
+  def process_access("private") do
```
geshuming: Single line functions can be simplified
```elixir
defp process_access("private"), do: "private"
defp process_access(_), do: "public"
```

martin-henz:
@geshuming is arguing that the presence and absence of `<password>` is enough as a distinction.

Simply add a password to protect the assessment, and remove the password to open it up.

What do you all think?

jiachen247:
Had a brilliant idea today, why dont we leverage `git submodule` instead of supporting two repos! seems like a much cleaner option.

The main more public repo can contain a submodule with stricter access control.

just have to edit a few commands and we should be good!
1. git pull to `git pull --recurse-submodules`
2. git clone to `git clone --recursive <>`

jiachen247:
@martin-henz @geshuming 

1) I still standby having the `practical` assessment type as it would make the spec more declarative. Doing without will definitely still work but I feel it communicates intent more clearly. We should just make a call and move on.

2) Yes thanks Shuming I totally agree that we should void the password once the assessment closes.

3) @martin-henz could you try to use git submodules with sws3012! would be nice to know that it works now that it has been merged in with #457

martin-henz:
OK, I've added a practical folder from a private repo as submodule. See PM in telegram @jiachen247

lib/cadet/assessments/assessments.ex
```diff
@@ -142,10 +142,18 @@ defmodule Cadet.Assessments do
...
-    case given_password do
-      nil -> {:error, {:forbidden, "Missing Password."}}
-      ^password -> assessment_with_questions_and_answers(assessment, user)
-      _ -> {:error, {:forbidden, "Invalid Password."}}
+    cond do
+      Timex.after?(Timex.now(), assessment.close_at) ->
+        assessment_with_questions_and_answers(assessment, user)
+
+      given_password == nil ->
+        {:error, {:forbidden, "Missing Password."}}
```
geshuming: Shouldn't this clause should be below the next clause `password == given_password`, else we will always fail on non-password protected assessments?

I think we should move it down just in case

jiachen247: it wouldnt reach here if it was a non password protected assessment! we overloaded the function like crazy opps. We can tell we didnt regress by the tests also!

lib/cadet/assessments/assessments.ex
```diff
@@ -121,15 +121,52 @@ defmodule Cadet.Assessments do
...
+
+      password == given_password ->
+        assessment_with_questions_and_answers(assessment, user)
+
+      true ->
+        {:error, {:forbidden, "Invalid Password."}}
+    end
+  end
+
+  def assessment_with_questions_and_answers(id, user = %User{}, password)
```
geshuming: I just realised, this is actually a bad way to check if an assessment is password protected or not. 

We should do something like

```elixir
def assessment_with_questions_and_answers(id, user = %User{}, _) when is_password_protected(id) do
  ...
end

def assessment_with_questions_and_answers(id, user = %User{}, password) do
  ...
end
```

jiachen247:
We should check if there is an associated submission and allow them in to prevent having to type in the password multiple times!

geshuming:
What if they want to view the questions first before attempting any?

jiachen247:
@geshuming I dont follow! the PE will be time constraint right? by checking for an existing submission we just make it easier for them if they accidentally exit the assessment workspace by pressing the back button or something.

lib/cadet/assessments/assessments.ex
jiachen247: `is_password_protected(id)` how would this work tho you will need to make a db call on id to check. long running functions dont make good guard clauses!

im actually using pattern matching to check if password protected
`def assessment_with_questions_and_answers(
        assessment = %Assessment{password: nil},
        user = %User{},
        nil
      ) do`

geshuming:
If students don't save an answer and decide to check all the questions first, there won't be a submission generated. As a result, the student will need to enter the password again just for navigating to a new question (as per current logic).

Ideally, we should create a submission when a user fetches the assessment anyway (to indicate the user has opened the assessment).

PR #406, opened 2019-07-02
Integration of Chatkit

flxffy:
Backend to supplement the usage of ChatKit. API can be found [here](https://pusher.com/docs/chatkit).

What's new:
1. `lib/cadet/chat/token.ex`: generate tokens for connection to Chatkit's services
2. `lib/cadet/chat/room.ex`: logic for creating rooms using Chatkit's API and storing room id in the db
2.1 Updated router to enable frontend retrieval of a token - for Chatkit's [TokenProvider](https://pusher.com/docs/chatkit/authentication#using-your-own-token-provider) in the frontend
3. `mix/tasks/users/chat.ex`: logic for creating users on Chatkit

Current Implementation:
- Rooms are currently being created at three points - when answers are created, when students edit their answers and when cadets access answer in the grading view. In the event Chatkit crashes, the last two points serve as failsafes to ensure comments get to the students.
  - Alternative: a script to create all chatrooms after Chatkit restores its services, if it crashes
- Making use of `comment` column in the Answer table to store room ids

Tests for Chatkit will be included by the end of the week once I figure out how ExVCR works 👯


flxffy:
> Hi, first thing I noticed is that there isn't a lot of tests...
> 
> Can we get some tests written so we can at least understand how to approach and use your features.

@geshuming There are no tests at the moment! They will be included once I figure out how ExVCR works. I'll ping the group once they are complete and ready for merging!

README.md
```diff
@@ -25,8 +25,9 @@ $ vim config/secrets.exs
     `luminus_redirect_url` are required for the application to properly authenticate with LumiNUS.
   - A valid `cs1101s_repository`, `cs1101s_rsa_key` is required for the application to 
     run with the `--updater` flag. Otherwise, the default values will suffice.
+  - A valid `instance_id`, `key_id` and `key_secret` are required to use ChatKit's services. Otherwise, the placeholder values can be left as they are.
 
-2. Install Elixir dependencies
+1. Install Elixir dependencies
```
geshuming: why is this 1

```suggestion
2. Install Elixir dependencies
```

lib/cadet/assessments/assessments.ex
```diff
@@ -5,6 +5,7 @@ defmodule Cadet.Assessments do
   """
   use Cadet, [:context, :display]
 
+  import Cadet.Chat.Room
```
geshuming: let's not import here. use `alias`

lib/cadet/assessments/assessments.ex
```diff
@@ -612,6 +618,11 @@ defmodule Cadet.Assessments do
       |> join(:inner, [a, ..., s], st in assoc(s, :student))
       |> preload([_, q, g, s, st], question: q, grader: g, submission: {s, student: st})
 
+    if role in @grading_roles or role in @see_all_submissions_roles do
+      # ChatKit - create chatrooms if they don't exist
+      create_rooms(Submission |> where(id: ^id) |> Repo.one())
```
geshuming: Why are you creating rooms in this function. This function is called when graders at the frontend views a submission.

lib/cadet/assessments/assessments.ex
```diff
@@ -744,11 +755,22 @@ defmodule Cadet.Assessments do
...
+           answer_changeset,
+           on_conflict: [set: [answer: get_change(answer_changeset, :answer)]],
+           conflict_target: [:submission_id, :question_id]
+         ) do
+      {:error, struct} ->
+        {:error, struct}
+
+      {:ok, struct} ->
+        # ChatKit - create chatrooms if they don't exist
+        create_rooms(submission)
```
geshuming: Why are you creating rooms here. This function is called when students save an answer on the frontend

geshuming:
Interestingly, we can already safely deprecate `comment` as the database stores it as "undefined"

capture2

lib/cadet/assessments/assessments.ex
flxffy: As we are now depending on an external service to host tutor comments, rooms are currently being created at three points - when answers are created, when students edit their answers and when cadets access answer in the grading view. In the event Chatkit crashes, the last two points serve as failsafes to ensure comments get to the students.

lib/cadet/assessments/assessments.ex
flxffy: I have answered this below, and have updated this PR's description with the current implementation.

> As we are now depending on an external service to host tutor comments, rooms are currently being created at three points - when answers are created, when students edit their answers and when cadets access answer in the grading view. In the event Chatkit crashes, the last two points serve as failsafes to ensure comments get to the students.

lib/cadet/assessments/assessments.ex
flxffy: Fixed with commit [3d3915b](https://github.com/source-academy/cadet/pull/406/commits/3d3915be3bd31d5b91b572ed2006f660ee2ed05c)

flxffy:
@geshuming With respect to the deprecation of `comment`, Martin would like to have this tested on real users during the summer workshop (14 July). Hence, in the interest of time, I will not be deprecating `comment` in this pull request as that would require changes on both front and back end.

geshuming:
@aihui ok, we can compromise with 1 small change. We will need to delete the `comment` attribute to prevent the case the `comment` field gets updated to "undefined" when avengers updates the grading info

https://github.com/source-academy/cadet/blob/20c99c43fafd0013171cb84b7c14d646f9a66a8c/lib/cadet/assessments/assessments.ex#L647-L655

```elixir
attrs
|> Map.put(attrs, "grader_id", grader_id)
|> Map.delete("comment")
```

I'm not very sure if the above suggestion will work, please test them out.

lib/cadet/assessments/assessments.ex
flxffy: An alternative to the current failsafes, if you would prefer, is to write a script to create all chatrooms after Chatkit restores its services, if it crashes. However, do note that the script needs to be run manually, and each request to their server averages 400-500ms.

flxffy:
@geshuming Thank you for pointing that out, I must have missed it earlier on. I will make the necessary changes and make a new commit tomorrow.

lib/cadet/assessments/assessments.ex
geshuming: I don't like this method. Can we try either creating a `mix task` that queries all `answer` in the database that is submitted but doesn't have a room, and generate the room key for them. Or we can try scheduling a task every 4 hours or every midnight to do the same. This way, we only need to generate the room key where it is really relevant to do so.

geshuming: oh i see you answered below

lib/cadet/assessments/assessments.ex
geshuming: > An alternative to the current failsafes, if you would prefer, is to write a script to create all chatrooms after Chatkit restores its services, if it crashes. The script will be scheduled to run daily to check for answers with an empty comment field (no chatroom), and make the request to create a new room.

Yes, i think this will be better.

flxffy: Changed with most recent [commit](https://github.com/source-academy/cadet/pull/406/commits/2b01b523689b3609b1621d56a9f4f291b29f7758). Script is scheduled to run at 1am, when hopefully there will be fewer students active online.

lib/cadet/assessments/assessments.ex
```diff
@@ -139,6 +140,12 @@ defmodule Cadet.Assessments do
         user = %User{role: role}
       ) do
     if Timex.after?(Timex.now(), assessment.open_at) or role in @open_all_assessment_roles do
+      # ChatKit - create chatrooms
+      Submission
+      |> where(assessment_id: ^id)
+      |> Repo.all()
+      |> Enum.each(fn submission -> Room.create_rooms(submission) end)
```
geshuming: Let's simplify this. There should only be one submission, so we should do

```elixir
 Submission
      |> where(assessment_id: ^id)
      |> Repo.one()
      |> &Room.create_rooms/1
```

lib/cadet/chat/room.ex
```diff
@@ -0,0 +1,89 @@
...
+        id: id,
+        student_id: student_id,
+        assessment_id: assessment_id
+      }) do
+    student = User |> where(id: ^student_id) |> Repo.one()
+
+    Answer
+    |> where(submission_id: ^id)
+    |> Repo.all()
+    |> Enum.filter(fn answer -> answer.comment == "" or answer.comment == nil end)
```
geshuming: Can be simplified

```suggestion
    |> Enum.filter(&(!&1.comment))
```

lib/cadet/chat/room.ex
```diff
@@ -0,0 +1,89 @@
...
+    |> Enum.each(fn answer ->
+      case create_room(assessment_id, answer.question_id, student) do
```
geshuming: Simplify

```elixir
 |> Enum.each(fn answer ->
  with {:ok, %{"id" => room_id}} <- create_room(assessment_id, answer.question_id, student) do
    # something
  end
end)

File: lib/cadet/chat/room.ex [hunk 8]
```diff
@@ -0,0 +1,89 @@
...
+      {:error, %HTTPoison.Error{reason: error}} ->
+        Logger.error("error: #{inspect(error, pretty: true)}")
+        {:error, nil}
+    end
+  end
+
+  defp get_staff_admin_user_ids do
+    User
+    |> Repo.all()
+    |> Enum.filter(fn user -> user.role == :staff or user.role == :admin end)
```
geshuming: Do the filtering at the database level.

```elixir
User
|> where([u], u.role in ^[:staff, :admin])
|> Repo.all()

File: lib/mix/tasks/chatkit_room.ex [hunk 9]
```diff
@@ -0,0 +1,26 @@
...
+  alias Cadet.Assessments.Submission
+  alias Cadet.Chat.Room
+
+  def run(_args) do
+    ensure_started(Repo, [])
+    HTTPoison.start()
+
+    Submission
+    |> Repo.all()
+    |> Enum.each(fn submission -> Room.create_rooms(submission) end)
```
geshuming: Filter the submissions at the database level. We shouldn't fetch the entire submission table every time.

```elixir
Submission
|> where(status: ^:submitted)
|> Repo.all()
|> Enum.each(fn submission -> Room.create_rooms(submission) end)
```

I think this will suffice, but there is definitely a better way to do this (filter submissions by whether the associated answers have comment)

lib/cadet/chat/room.ex
geshuming: Ignore this, just realised "" isnt falsey

lib/mix/tasks/chatkit_room.ex
flxffy: Can I check if graders are able to access submissions even if they are only attempted? Or are they only able to access the submissions that have already been submitted?

martin-henz: At the moment, graders only see student's attempts to solve an assessment after the attempt is submitted. Just attempting and saving currently does not give the graders (or any staff) access to the program.

lib/cadet/chat/room.ex
flxffy: Moved the room creation to `answer_question/3`. I think it makes more sense to have it after an answer is inserted or updated successfully.

lib/cadet/chat/room.ex
flxffy: Done

lib/mix/tasks/chatkit_room.ex
flxffy: Thank you for the clarification! Changes have been made accordingly.

lib/cadet/assessments/assessments.ex
flxffy: Moved the room creation to answer_question/3. I think it makes more sense to have it after an answer is inserted or updated successfully.

lib/cadet/chat/room.ex
```diff
@@ -0,0 +1,100 @@
...
+        "name" => "#{nusnet_id}_#{assessment_id}_Q#{question_id}",
+        "private" => true,
+        "user_ids" => get_staff_admin_user_ids() ++ [to_string(student_id)]
+      })
+
+    case HTTPoison.post(url, body, headers) do
+      {:ok, %HTTPoison.Response{body: body, status_code: 201}} ->
+        Poison.decode(body)
+
+      {:ok, _} ->
```
rrtheonlyone: Possible to check here for non-200 status and log it as well? e.g. for example if its a 400 status code stating "user not found".

I was testing this out locally and I realised my rooms were not being created due to user not found error. However, this was not logged and no indication was given to me.

lib/cadet/assessments/assessments.ex
```diff
@@ -351,8 +352,13 @@ defmodule Cadet.Assessments do
            {:is_open?, true} <- is_open?(question.assessment),
            {:ok, submission} <- find_or_create_submission(user, question.assessment),
            {:status, true} <- {:status, submission.status != :submitted},
-           {:ok, _} <- insert_or_update_answer(submission, question, raw_answer) do
+           {:ok, answer} <- insert_or_update_answer(submission, question, raw_answer) do
         update_submission_status(submission, question.assessment)
+
+        if answer.comment == "" or answer.comment == nil do
+          Room.create_rooms(submission, answer, user)
+        end
```
geshuming: Actually, can this be placed where a submission occurs?

lib/cadet/assessments/answer.ex
```diff
@@ -63,6 +63,11 @@ defmodule Cadet.Assessments.Answer do
     |> validate_xp_grade_adjustment_total()
   end
 
+  @spec comment_changeset(%__MODULE__{} | Ecto.Changeset.t(), map()) :: Ecto.Changeset.t()
+  def comment_changeset(answer, params) do
```
geshuming: Can we write a TODO to refactor this in the future. This is a very unsafe changeset function. So far, as long as you use it to change the `comment` it should be fine.

lib/cadet/chat/room.ex
```diff
@@ -0,0 +1,100 @@
...
+          student_id: student_id
+        }
+      ) do
+    user = User |> where(id: ^student_id) |> Repo.one()
+
+    Answer
+    |> where(submission_id: ^submission_id)
+    |> Repo.all()
+    |> Enum.filter(fn answer ->
+      answer.comment == "" or answer.comment == nil
```
geshuming: Can we remove the `answer.comment == ""` check, it doesn't seem to do anything. Let's keep it simple!

lib/cadet/assessments/assessments.ex
```diff
@@ -351,8 +352,13 @@ defmodule Cadet.Assessments do
            {:is_open?, true} <- is_open?(question.assessment),
            {:ok, submission} <- find_or_create_submission(user, question.assessment),
            {:status, true} <- {:status, submission.status != :submitted},
-           {:ok, _} <- insert_or_update_answer(submission, question, raw_answer) do
+           {:ok, answer} <- insert_or_update_answer(submission, question, raw_answer) do
         update_submission_status(submission, question.assessment)
+
+        if answer.comment == "" or answer.comment == nil do
```
geshuming: Remove `answer.comment == ""` check.

lib/mix/tasks/chatkit_room.ex
```diff
@@ -0,0 +1,28 @@
...
+  alias Cadet.Chat.Room
+
+  def run(_args) do
+    ensure_started(Repo, [])
+    HTTPoison.start()
+
+    Submission
+    |> where(status: ^:submitted)
+    |> Repo.all()
+    |> Enum.each(fn submission -> Room.create_rooms(submission) end)
```
geshuming:
```elixir
Submission
  |> join(:left, [s], a in assoc(s, :answers))
  |> preload([_, a], answers: a)
  |> where([_, a], is_nil(a.comment) or a.comment == "")
  |> Repo.all()
```

or since you have a function clause for answers, we can do

```elixir
Answer
  |> where([a], is_nil(a.comment) or a.comment == "")
  |> join(:left, [a], s in assoc(a, :submission))
  |> preload([_, s], submission: s)
  |> Repo.all()
```

lib/cadet/chat/room.ex
geshuming: What are the cases where `answer.comment == ""`?

geshuming:
> I prefer to use the NUS ID instead to create Users. This is just a suggestion/concern. @geshuming @flxffy do see if you want to change or not!

@rrtheonlyone It should be fine. Many other core components also use `user_id`. Normally user rows should be preserved, if a user drops out of the course for example we can just delete their authorization, and not the user row.

lib/cadet/assessments/assessments.ex
flxffy: May I clarify what you mean? Do you mean after submissions are finalised?

flxffy:
> Is there any way to check if a room key is valid? Currently it seems that we assume that `comment` is either `nil` or `VALID_KEY`

@geshuming The only way to verify a room id's validity is through a GET request to Chatkit to fetch room info. There's no documentation on how the room ids are being generated either.

lib/cadet/assessments/assessments.ex
geshuming: Yes. I think it is fine to leave the function here for now. But ultimately room creation should be done when a submission is submitted right?

flxffy: Sure. I'll place it [here](https://github.com/source-academy/cadet/blob/master/lib/cadet/jobs/autograder/grading_job.ex#L63) then? Since this is where all submissions flow through, whether it's submitted early or submitted by the autograder. Quite ugly tho :/

flxffy:
> I prefer to use the NUS ID instead to create Users. This is just a suggestion/concern. @geshuming @flxffy do see if you want to change or not!

@rrtheonlyone I used `user_id` because it's a unique identifier and a compulsory field. There's no `NOT NULL` constraint on the `nusnet_id` column so I'm going with `user_id` instead!

lib/cadet/chat/room.ex
flxffy: Logging added

lib/cadet/assessments/answer.ex
flxffy: Added a todo

lib/cadet/chat/room.ex
flxffy: I wasn't sure what `comment` was initialised with so I checked for empty strings. Removed now!

lib/cadet/assessments/assessments.ex
flxffy: Removed

lib/mix/tasks/chatkit_room.ex
flxffy: Done

flxffy:
Tests have been added for room creation.

PR #418, opened 2019-07-13
Integrate Sourcecast and Sourcereel components

jiayushe:
- Add table to store Sourcecast entries
- Support uploading and retrieving of binary data
- Relevant to source-academy/cadet-frontend#707


lib/cadet_web/controllers/sourcecast_controller.ex
```diff
@@ -0,0 +1,101 @@
...
+        |> put_status(status)
+        |> text(message)
+    end
+  end
+
+  swagger_path :index do
+    get("/sourcecast")
+    description("Lists all sourcecast files")
+    produces("application/json")
+    summary("Shows all files")
```
geshuming: Can we separate this action such that we do not send every audio file every time the frontend requests for this resource.

What should happen is that the `index` sends a list of sourcecast information, and maybe a `fetch` action for retrieving a specific sourcecast, along with the associated audio file.

geshuming: nvm i was under the impression that the audio file itself was sent

Issue #422, opened 2019-07-16
SWS3012 Missions not updating from repo


jiachen247:
Start and end times were updated on bitbucket and pull down to ec2b but never made it to the database. According to prof henz this was not the case last sem. To do root cause analysis to find out what caused this!

martin-henz:
Fixed by https://github.com/source-academy/cadet/pull/423

jiachen247:
@chrisjwelly thanks haha

martin-henz:
Note that this is fixed by https://github.com/source-academy/cadet/pull/423
which is merged AND already deployed in the current cadet backend.

PR #423, opened 2019-07-16
update external libs for 1920

jiachen247:
updated extenal lib enum on the backend to be up to date with the new frontend libs

image


jiachen247:
Resolves #422

martin-henz:
Note that this is merged AND already deployed in the current cadet backend.

PR #437, opened 2019-07-19
Added clause for questions with no testcases

### LINKED ISSUE #434: Don't send a request if there are no testcases
https://github.com/source-academy/cadet/blob/5980abb53642795c4da463903070aee83e4265ce/lib/cadet/jobs/autograder/lambda_worker.ex#L20

geshuming:
Fix #434

Check if there are testcases before invoking lambda function. Will not invoke if there are no testcases.


lib/cadet/jobs/autograder/lambda_worker.ex
```diff
@@ -18,16 +18,20 @@ defmodule Cadet.Autograder.LambdaWorker do
   storage job.
   """
   def perform(params = %{answer: answer = %Answer{}, question: %Question{}}) do
-    lambda_params = build_request_params(params)
+    with lambda_params <- build_request_params(params),
+         {:hasTestcases?, true} <- {:hasTestcases?, !Enum.empty?(lambda_params.testcases)} do
```
jiachen247: why suddenly camel case here? `contains_testcases?` seems like a better fit haha

jiachen247: Also, i think an if else or a case here will be cleaner leh the function is quite unreadable! With should be used for multi step transaction! defo prefer if else of cond if youre just checking one thing!

geshuming: Thanks for the suggestion

PR #438, opened 2019-07-19
Allow graders to grade before autograding

### LINKED ISSUE #436: Autograding should not alter manual grading
https://github.com/source-academy/cadet/blob/5980abb53642795c4da463903070aee83e4265ce/lib/cadet/jobs/autograder/result_store_worker.ex#L60

geshuming:
Fix #436 

the adjustment is calculated and updated such that the total grade/xp should be 0 <= total grade/xp <= max, hence preserving the previous validation.

### DIFF
--- a/lib/cadet/assessments/answer.ex
+++ b/lib/cadet/assessments/answer.ex
@@ -80,25 +80,38 @@ defmodule Cadet.Assessments.Answer do
     with {:question_id, question_id} when is_ecto_id(question_id) <-
            {:question_id, answer.question_id},
          {:question, %{max_grade: max_grade, max_xp: max_xp}} <-
-           {:question, Repo.get(Question, question_id)},
-         {:total_grade, true} <- {:total_grade, total_grade >= 0 and total_grade <= max_grade},
-         {:total_xp, true} <- {:total_xp, total_xp >= 0 and total_xp <= max_xp} do
+           {:question, Repo.get(Question, question_id)} do
       changeset
+      |> put_change(
+        :adjustment,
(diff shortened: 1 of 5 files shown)


jiachen247:
I dont think the solution here is just not touch the adjustment. This might throw the total grade outside the acceptable range but i think more than that might be super misleading for the student.

Imagine he made mistakes and got 50/100 manual grading. autograder runs and gives him 50/100 as well. Now from his point of view his computed grade is 100/100. The grader would be blind to this as well.

jiachen247:
Of the top of my head I can think of two solutions,
1. We make the grader a little smarter and computes and the new adjustment value such that the manual grade is maintained.
2. if manually graded we dont autograde anymore

jiachen247:
In addition, keeping missions due at 2359 and triggering grading jobs at 1201 seems to be a good move as well.

geshuming:
I think we can go for the 1st suggestion. The 2nd suggestion is not really viable, because we might want to rely on automated testing. 

I guess we can compute a new adjustment such that the total grade stays within the bounds. Will look into it

jiachen247:
Thanks man! Solid!

geshuming:
Oh i thought we just need to preserve the bounds.

> if autograder gives a high grade than the manaul grader then ignore the autograding
> if its less the set the grade to the auto grader and deduct that from the adjusment

I don't think this is a good idea. Manual grading is separate from autograding. We might look at assessments where autograding takes up 9 marks, and manual grading takes up 3 marks. Graders should not take up the role of auto graders, but should still be free to double check on the autograding component and adjust accordingly.

geshuming:
Let's preserve the final grade based on adjustment

PR #440, opened 2019-07-19
Configure Lambda to use Node 10.x

geshuming:


jiachen247:
Can check with the slang guys @open

geshuming:
I'm matching the front-end's node version

PR #451, opened 2019-07-25
Use public chatkit chatrooms

### LINKED ISSUE #448: Log error to sentry when creating rooms for chatkit
Currently failure to generate rooms just fail quietly

geshuming:
Resolves #448 
Resolves https://github.com/source-academy/cadet-frontend/issues/750
Resolves https://github.com/source-academy/cadet-frontend/issues/749


jiachen247:
Could you briefly describe how your PR solves the issue thanks!

geshuming:
Due to [Pusher's response on the user limit](https://github.com/source-academy/cadet-frontend/issues/750#issuecomment-513755995), we will send a request for every 10 users to add the users into the room. 

But, in hindsight, I think the public chatroom solution might be better...

geshuming:
I will modify to use `public` chatrooms instead

geshuming:
Modified to use public instead of private chatrooms now

PR #458, opened 2019-07-26
allow mutation of open assessment

### LINKED ISSUE #435: Mutating Open Assessments
Currently, missions that are open are no longer updated when the updater run. Im guessing this rule was created to ensure that missions are not mutated after it is open.

However, over the last sems, mistakes with missions do happen and when they do they are labeled with the OBSOLETE in the title and a new one is created in place.

This creates some issues. Its confusing for students keeping track of the right assessments and two it makes it impossible to track progress cause the number of missions keep increasing. See https://github.com/source-academy/cadet/issues/129

To solve this, I propose we make the check a little smarter. I suggest we allow alterations to open missions provided no questions are added or removed. (This messes with the foreign keys).

jiachen247:
Resolves #435

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -246,7 +246,7 @@ defmodule Cadet.Assessments do
       Multi.insert_or_update(
         Multi.new(),
         :assessment,
-        insert_or_update_assessment_changeset(assessment_params)
+        insert_or_update_assessment_changeset(assessment_params, questions_params)
       )
 
     questions_params
@@ -266,8 +266,10 @@ defmodule Cadet.Assessments do
     |> Repo.transaction()
   end
 
-  @spec insert_or_update_assessment_changeset(map()) :: Ecto.Changeset.t()
-  defp insert_or_update_assessment_changeset(params = %{number: number}) do
+  @spec insert_or_update_assessment_changeset(map(), map()) :: Ecto.Changeset.t()
+  defp insert_or_update_assessment_changeset(params = %{number: number}, questions_params) do
+    new_question_count = Enum.count(questions_params)
+
     Assessment
     |> where(number: ^number)
     |> Repo.one()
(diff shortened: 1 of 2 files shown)


lib/cadet/assessments/assessments.ex
```diff
@@ -276,18 +278,24 @@ defmodule Cadet.Assessments do
...
+        %{id: assessment_id} = assessment
 
+        question_count =
+          Question |> where(assessment_id: ^assessment_id) |> Repo.aggregate(:count, :id)
+
+        if Timex.after?(assessment.open_at, Timex.now()) or question_count == new_question_count do
+          # Delete all existing questions
           Question
           |> where(assessment_id: ^assessment_id)
           |> Repo.delete_all()
```
geshuming: I don't think this works for mutation though. What happens if there are existing answers? How will we map the new question to the existing answers. I believe we currently use question assoc when we deal with answers in some parts of the code.

geshuming: Another issue is that... if I'm not wrong, because of the hourly assessment updater task, we are actually deleting and inserting new questions every hour. I can't believe I am seeing question ids in the 3000s in the current SWS version. 

I think this assessment updating logic needs to be reworked.

jiachen247: I feel like there has to be a reason why they chose to delete and re insert! let me check with them haha

jiachen247:
Okay this seems like a harder problem to solve due to the way the assessments are updated! It would take a bigger refactor to implement such functionality. KIV for now!

martin-henz:
> Okay this seems like a harder problem to solve due to the way the assessments are updated! It would take a bigger refactor to implement such functionality. KIV for now!

OK, no problem. We will keep the workflow for the current semester: Mission authors will use the live system to preview the missions. The mission authors will make sure that they set the release date to a default value, for example December 1, 2019. Before the mission is open, staff can freely test. Once tested, the mission author will set the release date to the intended date.

jiachen247:
Yesss thanks prof!! This really deserves another look in winter I think its something that could really improve the assessment workflow if implement properly!!

PR #494, opened 2019-09-12
Fix test randomly failing due to constraint error

geshuming:
# Issue fixed in this PR

image

The above test has a 10% chance to fail due to Paths and Practicals sharing the same first letter.


jiachen247:
Could you elaborate on this issue a little more! what does the test do and how come it fails when there share the same first letter? @geshuming

geshuming:
@jiachen247 there is a unique constraint on the number field. The test randomly generates a string "<letter><number" for each type of assessment where the letter is the first letter of the type, and the number is random between 1 to 10.

This pr resolves the issue by replacing <letter> with the full name of the type of assessment.

PR #514, opened 2019-10-06
Fix incorrect license in README

chrisgzf:
The license for this project was updated in https://github.com/source-academy/cadet/pull/420, but the README was not updated to reflect that.


jiachen247:
Closed with #515

PR #517, opened 2019-10-13
Extend guardian access token expiry timeout

jiachen247:
@Aulud can you link the issues here I cant seem to find them!

Increased expiry timeout of the access token from 60 to 180 hours which should be more than sufficient for the PE. Quick and dirty fix to look into a cleaner solution :-)


Aulud:
> @Aulud can you link the issues here I cant seem to find them!
> 
> Increased expiry timeout of the access token from 60 to 180 hours which should be more than sufficient for the PE. Quick and dirty fix to look into a cleaner solution :-)

Woah woah 180 hours 🤣 it's 180 minutes haha

The issue's at https://github.com/source-academy/cadet-frontend/issues/934.

jiachen247:
I mean the refresh token is already on the client-side (contrary to best practices) so I see no harm just merging it now!

PR #580, opened 2020-04-13
Modules and Webhooks Support

Hou-Rui:
Add support for the front-end to fetch modules from the back-end;
Add Webhooks support in the back-end.


martin-henz:
This PR decreases test coverage by quite a bit. Can you try to maintain the coverage, by writing test cases?

BoAi01:
Test cases added. Coverage can maintain 91% (0.09% lower).

lib/cadet_web/controllers/module_controller.ex
```diff
@@ -0,0 +1,46 @@
...
+    Synchronize local module files witn GitHub repository via WebHook
+    and repond HTTP resquest from frontend
+    """
+  
+    use CadetWeb, :controller
+  
+    use PhoenixSwagger
+  
+    def handle(conn, _payload) do
+      repo = Git.new("priv/module_static")
```
martin-henz, line 12: wondering about the choice of name for the folder. This is the destination of the webhook, right? So priv/module_static gets populated by the modules from the modules repo, right? In that case, what is "static" about it?

priv/module_static/lib/binary_tree.js
```diff
@@ -0,0 +1,90 @@
+// tree.js: Binary Tree abstraction for M5
```
martin-henz, line 1: Having libraries explicitly included here will violate the "single source of truth" principle. We would have two versions and it wouldn't be clear what version is available in what situation.

priv/module_static/lib/runes.js
```diff
@@ -0,0 +1,6463 @@
+(function (exports) {
```
martin-henz, line 1: same as for binary_tree.js

lib/cadet_web/controllers/module_controller.ex
BoAi01: Okay, we can change the name of the destination folder, what about putting the module files under priv/modules?

priv/module_static/lib/binary_tree.js
BoAi01: Okay, I have added the files under priv/modules to .gitignore.

priv/module_static/lib/runes.js
BoAi01: Same as above. Modules files are added to .gitignore.

priv/module_static/lib/runes.js
```diff
@@ -6407,7 +6413,7 @@
   getReadyWebGLForCanvas('3d');
 
   exports.anaglyph = anaglyph;
-  exports.animate = animate;
+  exports.animated_show = animated_show;
```
travisryte: How can I test this?

Hou-Rui: You need a journey-frontend to test this. The whole new lib/runes requires to be run on the journey-frontend.

angelsl:
Going to close this as we are thinking of a backend-independent module system.

PR #583, opened 2020-04-14
Modules and Webhooks Support (test cases added)

BoAi01:
Add support for the front-end to fetch modules from the back-end;
Add Webhooks support in the back-end.
Add test cases for ModuleController.


BoAi01:
Sorry I should merge to sa_2021.

PR #638, opened 2020-07-20
Add remote execution endpoints

angelsl:


angelsl:
Forgot to add Swagger documentation. Can review the rest first I guess.

lib/cadet/devices/devices.ex
```diff
@@ -0,0 +1,348 @@
...
+        )
+
+      # ExAws includes the session token in the signed payload and doesn't allow
+      # you not to do so. Some AWS services require it to be in the signed
+      # payload, some don't. This one doesn't, so.. we manually append the
+      # security token. *sigh*
+      {:ok,
+       %{
+         endpoint:
+           signed_url <> "&X-Amz-Security-Token=" <> URI.encode_www_form(creds.session_token),
```
Aulud, line 219: I suppose we could use string interpolation here instead of concatenation.

lib/cadet/helpers/shared_helper.ex
```diff
@@ -20,12 +20,24 @@ defmodule Cadet.SharedHelper do
     end)
   end
 
+  @doc """
```
Aulud, line 23: Nice, we have more documentation now.

lib/cadet_web/controllers/devices_controller.ex
```diff
@@ -0,0 +1,118 @@
...
+        send_resp(conn, :bad_request, DisplayHelper.full_error_messages(changeset))
+    end
+  end
+
+  def edit(conn, %{"id" => device_id, "title" => title}) do
+    with {:get_registration, registration} when not is_nil(registration) <-
+           {:get_registration,
+            Devices.get_user_registration(conn.assigns.current_user, device_id)},
+         {:rename, {:ok, _}} <- {:rename, Devices.rename_registration(registration, title)} do
+      send_resp(conn, :no_content, "")
```
Aulud, line 31: Nice - I think there are a few other endpoints (not in this PR) that should use 204 in place of 200 with a redundant "OK" text string.

lib/cadet_web/views/devices_view.ex
```diff
@@ -0,0 +1,16 @@
...
+    render_many(registrations, CadetWeb.DevicesView, "show.json", as: :registration)
+  end
+
+  def render("show.json", %{registration: registration}) do
+    %{
+      id: registration.id,
+      title: registration.title,
+      type: registration.device.type,
+      secret: registration.device.secret
+    }
```
Aulud, line 14: Whilst most of the other views use `transform_map_for_view` instead, this actually ends up much cleaner overall, so its fine.

lib/cadet/devices/devices.ex
```diff
@@ -0,0 +1,348 @@
...
+  alias Cadet.Accounts.User
+  alias Cadet.Devices.{Device, DeviceRegistration}
+  alias ExAws.STS
+
+  @spec get_user_registrations(integer | %User{}) :: [%DeviceRegistration{}]
+  def get_user_registrations(%User{id: user_id}) do
+    get_user_registrations(user_id)
+  end
+
+  def get_user_registrations(user_id) when is_integer(user_id) or is_binary(user_id) do
```
Aulud, line 19: Seems a bit peculiar `get_user_registrations/1` here accepts a binary `user_id` whilst `get_user_registration/2` and `register/4` below does not.

test/cadet/devices/devices_test.exs
```diff
@@ -0,0 +1,228 @@
...
+    use_cassette "aws/devices_get_ws_endpoint#3", custom: true do
+      assert {:error, _} = Devices.get_device_ws_endpoint(device.id, user)
+    end
+  end
+
+  defp assert_submaps_eq(expected, actual, fields) do
+    assert length(expected) == length(actual)
+
+    Enum.map(Enum.zip([expected, actual]), fn {e, a} -> assert_submap_eq(e, a, fields) end)
+  end
```
Aulud, line 208: This looks like it's essentially checking for structural equality - not sure if Elixir's `Map.equal?/2` does this. That aside, this seems generally useful enough to be moved into a file for test helpers.

PR #713, opened 2021-01-27
Update `/assets` endpoint

bnjmnt4n:
I've created a `v2` branch, which me and @qreoct will use over the course of the semester to stage our changes.

This PR removes the role validation from the `Assets` model, since models should not be concerned with validation, and removes the `user` parameter from the function definitions. It also updates the routes to shift the `/v1/assets` endpoint into `/v2/admin/assets` and performs validation in the router.


test/cadet_web/admin_controllers/admin_assets_controller.ex
```diff
@@ -36,14 +36,14 @@ defmodule CadetWeb.AssetsControllerTest do
     @tag authenticate: :student
     test "GET /assets/:foldername", %{conn: conn} do
       conn = get(conn, build_url("testFolder"), %{})
-      assert response(conn, 403) =~ "User not allowed to manage assets"
+      assert response(conn, 403) =~ "Forbidden"
```
bnjmnt4n, line 39: We've lost the more informative error message, but I think there's a fair trade-off in terms of improving code de-duplication.

lib/cadet/assets/assets.ex
```diff
@@ -35,19 +32,17 @@ defmodule Cadet.Assets.Assets do
     end
   end
 
-  def list_assets(folder_name, user) do
-    with :ok <- validate_role(user.role),
-         :ok <- validate_folder_name(folder_name) do
+  def list_assets(folder_name) do
+    with :ok <- validate_folder_name(folder_name) do
```
bnjmnt4n, line 36: In terms of code style, is this fine or should I use a `case` instead of a `with` statement with only 1 match sequence? (as someone who is new to Elixir :pray: )

angelsl: Using `case` is better in this case, yes

angelsl (commented):
Looks good to me aside from `with`/`case` which is mostly a minor stylistic thing.

Actually, I was thinking that you might keep the v1 routes around until the v2 routes were ready, but honestly this is fine too.

PR #722, opened 2021-02-17
Update /settings endpoint

qreoct:
Shift `PUT /settings/sublanguage` to `PUT /admin/settings/sublanguage`, since the latter is only allowed for authorized users, to align with our v2 API design guidelines.


bnjmnt4n (approved):
LGTM!

PR #726, opened 2021-02-21
Fix incentives entities' field names

jonas-chow:
Fixes some problems to do with associations between achievements and goals/other achievements.
Also renames max_xp to maxXp, and prerequisiteIds to prerequisiteUuids to conform with the frontend naming.

### DIFF
--- a/lib/cadet/incentives/goals.ex
+++ b/lib/cadet/incentives/goals.ex
@@ -20,12 +20,29 @@ defmodule Cadet.Incentives.Goals do
 
   @doc """
   Returns goals with user progress.
+  Renames max_xp to maxXp in the meta
   """
   def get_with_progress(%User{id: user_id}) do
     Goal
     |> join(:left, [g], p in assoc(g, :progress), on: p.user_id == ^user_id)
     |> preload([g, p], progress: p)
     |> Repo.all()
+    |> Enum.map(&change_meta(&1))
+  end
+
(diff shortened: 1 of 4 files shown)


lib/cadet/incentives/goals.ex
```diff
@@ -20,12 +20,29 @@ defmodule Cadet.Incentives.Goals do
...
+      |> Map.put("maxXp", Map.fetch!(meta, "max_xp"))
+      |> Map.delete("max_xp")
+    else
+      meta
+    end
+  end
+
+  defp change_meta(goal) do
+    goal
+    |> Map.put(:meta, change_max_xp_name(goal.meta))
```
angelsl, line 45: This should not be here. Meta is entirely freeform data set by the frontend, so the frontend should just set it in the correct format that it wants when creating the goal..

lib/cadet_web/admin_controllers/admin_achievements_controller.ex
```diff
@@ -5,6 +5,7 @@ defmodule CadetWeb.AdminAchievementsController do
 
   alias Cadet.Incentives.Achievements
 
+  @spec bulk_update(Plug.Conn.t(), map) :: Plug.Conn.t()
```
angelsl, line 8: Please remove this type hint, it's not needed.

angelsl (changes requested):

angelsl (approved):

PR #727, opened 2021-02-21
Update `/grading` endpoint

qreoct:
Move the Grading Controller to behind /admin endpoint for v2 API consistency. 

Some (not all) of role validation was moved out of the `Assessments` Model to the router. The methods that are required by the new AdminGrading Controller have all been fixed, but there are some test cases relating to Assessments Controller which are not rectified in this PR. @bnjmnt4n, let me know if the changes I've made aligns with your Assessments changes.

The specific 3 tests that don't pass are involved with Students making API calls in AssessmentController, where the Students are directly authorized (with 200 response) when they shouldn't be. The tests fail since I did not move the /assessments endpoints to be role-validated in the router in this PR.

### DIFF
--- a/lib/cadet_web/admin_controllers/admin_grading_controller.ex
+++ b/lib/cadet_web/admin_controllers/admin_grading_controller.ex
@@ -1,4 +1,4 @@
-defmodule CadetWeb.GradingController do
+defmodule CadetWeb.AdminGradingController do
   use CadetWeb, :controller
   use PhoenixSwagger
 
@@ -15,11 +15,6 @@ defmodule CadetWeb.GradingController do
         |> put_status(:ok)
         |> put_resp_content_type("application/json")
         |> text(submissions)
-
-      {:error, {status, message}} ->
-        conn
-        |> put_status(status)
-        |> text(message)
     end
   end
 
@@ -28,9 +23,7 @@ defmodule CadetWeb.GradingController do
   end
 
   def show(conn, %{"submissionid" => submission_id}) when is_ecto_id(submission_id) do
-    user = conn.assigns[:current_user]
-
-    case Assessments.get_answers_in_submission(submission_id, user) do
+    case Assessments.get_answers_in_submission(submission_id) do
       {:ok, answers} ->
         render(conn, "show.json", answers: answers)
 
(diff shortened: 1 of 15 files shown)


lib/cadet_web/admin_controllers/admin_grading_controller.ex
```diff
@@ -129,16 +122,9 @@ defmodule CadetWeb.GradingController do
   end
 
   def grading_summary(conn, _params) do
-    user = conn.assigns[:current_user]
-
-    case Assessments.get_group_grading_summary(user) do
+    case Assessments.get_group_grading_summary() do
```
qreoct, line 125: Example where user role validation is moved out of the Model to the router.

We lose a descriptive error message, and get a 403 Forbidden when unauthorized users access these endpoints.

lib/cadet_web/admin_controllers/admin_grading_controller.ex
```diff
@@ -249,6 +235,20 @@ defmodule CadetWeb.GradingController do
     response(401, "Unauthorised")
   end
 
+  swagger_path :grading_summary do
```
qreoct, line 238: Added Swagger documentation for the `/grading/summary` endpoint, which was previously undocumented.

test/cadet_web/admin_controllers/admin_grading_controller_test.exs
```diff
@@ -49,45 +49,45 @@ defmodule CadetWeb.GradingControllerTest do
...
     test "unauthorized", %{conn: conn} do
       conn = post(conn, build_url(1, 3), %{"grading" => %{}})
-      assert response(conn, 401) =~ "User is not permitted to grade."
+      assert response(conn, 403) =~ "Forbidden"
     end
 
     @tag authenticate: :student
     test "missing parameter", %{conn: conn} do
       conn = post(conn, build_url(1, 3), %{})
-      assert response(conn, 400) =~ "Missing parameter"
```
qreoct, line 82: Now we give a 403 Forbidden since a student would never be authorized to make this API call anyway.

bnjmnt4n (approved):
:+1: Will merge into `v2` and fix the errors by updating `Assessments`.

PR #747, opened 2021-04-16
Update achievements structure

jonas-chow:
- Implements endpoint to update goal progresses
- Achievements have a few new fields: xp, variable_xp, 
- Goals have a field renamed: max_xp -> target_count
- GoalProgress have a field renamed: xp -> count
- GoalsToAchievements is now a many-to-many association


lib/cadet_web/controllers/incentives_controller.ex
```diff
@@ -15,6 +15,26 @@ defmodule CadetWeb.IncentivesController do
     )
   end
 
+  def update_progress(conn, %{"uuid" => uuid, "userid" => userid, "progress" => progress}) do
+    progress
+    |> json_to_progress(uuid, userid)
+    |> Goals.upsert_progress(uuid, userid)
+    |> handle_standard_result(conn)
+  end
```
angelsl, line 23: This doesn't do any validation on `userid`, so students could update other students' goal progress.

lib/cadet_web/router.ex
```diff
@@ -51,6 +51,7 @@ defmodule CadetWeb.Router do
     resources("/sourcecast", SourcecastController, only: [:create, :delete])
 
     get("/achievements", IncentivesController, :index_achievements)
+    post("/achievements/goals/:uuid/:userid", IncentivesController, :update_progress)
```
angelsl, line 54: Have you cleared this with the backend team? This is not a RESTful endpoint.

lib/cadet_web/controllers/incentives_controller.ex
angelsl: Also, missing Swagger documentation for this new endpoint.

angelsl (changes requested):
Some minor comments, and one major comment, as I mentioned in the frontend PR:

The endpoint for updating goal progress needs to be split into one for staff (that can update any student's goal progress), and one for students only to log event occurrences. The XP and achievement count goals should be handled in the backend.

angelsl (approved):

PR #771, opened 2021-05-15
Fix bug with leaderboard contest_number field not found

kevin9foong:
Fix a bug with the contest_number being null when loading contest entries


angelsl (approved):
Sometimes I wish we were using a strongly-typed language..

PR #775, opened 2021-06-08
Multitenant backend

YaleChen299:
This is the initial changes for the backend, we mostly covered accounts and courses context, we will move on to assessment.
all test under /courses and /accounts are passing.
All the tests should be passing.

(seeds.exs is working partially, but migration from old db id not working, please use `MIX_ENV=test mix ecto.reset` before testing)

PR #779, opened 2021-06-19
Added auth provider namespacing, and Google and GitHub auth

chownces:
Note that this will currently fail as `Accounts.sign_in` is configured to `get_role` from the authentication provider when signing in for the first time. This will be a non-issue in the future as roles will no longer be tied to users, but the course that the user is in.

### DIFF
--- a/lib/cadet/auth/providers/github.ex
+++ b/lib/cadet/auth/providers/github.ex
@@ -32,7 +32,7 @@ defmodule Cadet.Auth.Providers.GitHub do
            {:token, HTTPoison.post(@token_url, token_query, token_headers)},
          {:token_response, %{"access_token" => token}} <- {:token_response, Jason.decode!(body)},
          {:user, {:ok, %{"login" => username}}} <- {:user, api_call(@user_api, token)} do
-      {:ok, %{token: token, username: username}}
+      {:ok, %{token: token, username: Provider.namespace(username, "github")}}
     else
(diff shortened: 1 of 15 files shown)


lib/cadet/auth/providers/github.ex
```diff
@@ -0,0 +1,76 @@
...
+  @user_api "https://api.github.com/user"
+
+  @spec authorise(config(), Provider.code(), Provider.client_id(), Provider.redirect_uri()) ::
+          {:ok, %{token: Provider.token(), username: String.t()}}
+          | {:error, Provider.error(), String.t()}
+  def authorise(config, code, client_id, redirect_uri) do
+    token_query =
+      URI.encode_query(%{
+        client_id: client_id,
+        client_secret: config.client_secret,
```
angelsl, line 21: Should either

1. have a map of client_id => client_secret in the provider config (preferred); or
2. have client_id in the provider config and check that the client_id matches the one in the provider config

LumiNUS is a bit special because NUS ADFS is not configured to use any client secret.

The reason we want to accept multiple client IDs is so that we can use frontends hosted in multiple locations with a single backend. (Each client ID typically has a different (set of) accepted redirect URL(s).)

chownces:
Have already merged this into the main multitenant PR. Will just leave this here in the meantime for reference as the other PR is getting really huge

chownces:
Duplicate. Refer to #775

PR #781, opened 2021-06-26
Added update_role and delete_user endpoints

chownces:
`delete_user` might be a duplicate of `delete_record` in `accounts/course_registrations.ex`. feel free to copy/ merge these changes and tests over to `accounts/course_registrations.ex` where you see fit

tested these 2 endpoints on frontend alr


YaleChen299 (approved):
Generally looks good! To be refactored later

PR #820, opened 2021-09-03
fix insert submission_votes to a course

YaleChen299:
This PR serves to fix various loopholes regarding contest.


YaleChen299:
Note to self, This temporary fix cause a decrease in coverage in assessment_helper contest_leaderboard. Will fix that after leaderboard hiding logic is implemented.

PR #823, opened 2021-09-08
Remove rank from contest voting

lizchow:
Remove the concept of rank from contest voting. The scoring for voting is now 1(worst) to 10 (best). 

Should be tested with the frontend PR.

### DIFF
--- a/lib/cadet_web/views/assessments_helpers.ex
+++ b/lib/cadet_web/views/assessments_helpers.ex
@@ -129,7 +129,7 @@ defmodule CadetWeb.AssessmentsHelpers do
     transform_map_for_view(entry, %{
       submission_id: :submission_id,
       answer: :answer,
-      rank: :rank
+      score: :score
     })
   end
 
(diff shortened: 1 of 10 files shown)


lib/cadet_web/views/assessments_helpers.ex
```diff
@@ -140,7 +140,7 @@ defmodule CadetWeb.AssessmentsHelpers do
         answer: :answer,
         student_name: :student_name
       }),
-      "score",
+      "relative_score",
```
YaleChen299, line 143: will this be confusing with the score? Is this relative score the computed final score on the leaderboard? Maybe we can name it `final_score` or something along those line?

YaleChen299 (commented):
Generally looks good to me! Just one minor issue on naming that may affect the frontend side as well.

lib/cadet_web/views/assessments_helpers.ex
lizchow: ok

YaleChen299:
percentage dropped due to removed parts, no new uncovered lines.

PR #824, opened 2021-09-17
Update Elixir and OTP versions

angelsl:


angelsl:
Not sure why the coverage changed...

PR #833, opened 2021-10-14
Allow questions to specify Source variant and execution time

angelsl:


martin-henz:
Let's remember to update this:
https://about.sourceacademy.org/educator/assessment/#deployment

The variant should contain allowable combinations, such as
interpreter: 2
variant: "lazy"

PR #859, opened 2022-03-01
Add Total XP Calculation

ShenyiCui:
## Changes
- New Incentive Controller function `calculate_totalxp()`
- New Achievement function `get_total_xp()`
- Calling the API endpoint `calculate_totalxp` will return a value denoting the current total XP for the current user.

## Testing 
- Navigate to `dev.exs` and provide your local database credentials
- Open API link: [Postman Testing](https://documenter.getpostman.com/view/16373026/UVksLZYw#6c1cb230-390b-4c61-a799-adc391a061a6)
- Retrieve an `authorisation token`
- Set your `course_id`
- Run the API call **Get Total XP**

**Expected Result**
Screenshot 2022-03-02 at 1 02 32 AM


config/dev.exs
```diff
@@ -51,9 +51,9 @@ config :phoenix, :stacktrace_depth, 20
 
 # Configure your database
 config :cadet, Cadet.Repo,
-  username: "postgres",
-  password: "postgres",
-  database: "cadet_dev",
+  username: "###",
+  password: "###",
+  database: "###",
```
angelsl, line 56: Please don't change this

lib/cadet/incentives/achievements.ex
```diff
@@ -21,7 +21,43 @@ defmodule Cadet.Incentives.Achievements do
...
+    end
+
+    acc_achievements = fn (achievement_item, acc) ->
+      acc + acc_goals.(achievement_item.goals) * achievement_item.xp
+    end
+
+    Enum.reduce(achievements, 0, acc_achievements)
+  end
+
+  @spec upsert(map()) :: {:ok, %Achievement{}} | {:error, {:bad_request, String.t()}}
```
angelsl, line 60: Did you accidentally change this line? `Achievement.t()` became `%Achievement{}`

lib/cadet/incentives/achievements.ex
```diff
@@ -21,7 +21,43 @@ defmodule Cadet.Incentives.Achievements do
     |> Repo.all()
   end
 
-  @spec upsert(map()) :: {:ok, Achievement.t()} | {:error, {:bad_request, String.t()}}
+  @doc """
+  Returns all achievements with goals and progress.
+  # To Do
+  Run elixir tests for edge cases etc.
```
angelsl, line 27: Please don't leave todos in doc comments.

lib/cadet/incentives/achievements.ex
```diff
@@ -21,7 +21,43 @@ defmodule Cadet.Incentives.Achievements do
     |> Repo.all()
   end
 
-  @spec upsert(map()) :: {:ok, Achievement.t()} | {:error, {:bad_request, String.t()}}
+  @doc """
+  Returns all achievements with goals and progress.
```
angelsl, line 25: This doesn't seem to match the function's actual behaviour

mix.lock
```diff
@@ -5,26 +5,26 @@
   "blankable": {:hex, :blankable, "1.0.0", "89ab564a63c55af117e115144e3b3b57eb53ad43ba0f15553357eb283e0ed425", [:mix], [], "hexpm", "7cf11aac0e44f4eedbee0c15c1d37d94c090cb72a8d9fddf9f7aec30f9278899"},
   "bunt": {:hex, :bunt, "0.2.0", "951c6e801e8b1d2cbe58ebbd3e616a869061ddadcc4863d0a2182541acae9a38", [:mix], [], "hexpm", "7af5c7e09fe1d40f76c8e4f9dd2be7cebd83909f31fee7cd0e9eadc567da8353"},
   "bypass": {:hex, :bypass, "2.1.0", "909782781bf8e20ee86a9cabde36b259d44af8b9f38756173e8f5e2e1fabb9b1", [:mix], [{:plug, "~> 1.7", [hex: :plug, repo: "hexpm", optional: false]}, {:plug_cowboy, "~> 2.0", [hex: :plug_cowboy, repo: "hexpm", optional: false]}, {:ranch, "~> 1.3", [hex: :ranch, repo: "hexpm", optional: false]}], "hexpm", "d9b5df8fa5b7a6efa08384e9bbecfe4ce61c77d28a4282f79e02f1ef78d96b80"},
-  "certifi": {:hex, :certifi, "2.8.0", "d4fb0a6bb20b7c9c3643e22507e42f356ac090a1dcea9ab99e27e0376d695eba", [:rebar3], [], "hexpm", "6ac7efc1c6f8600b08d625292d4bbf584e14847ce1b6b5c44d983d273e1097ea"},
```
angelsl, line 8: Please undo the changes to mix.lock; if you want to update dependencies, do that in a separate PR

lib/cadet_web/router.ex
```diff
@@ -79,6 +79,7 @@ defmodule CadetWeb.Router do
     post("/assessments/question/:questionid/answer", AnswerController, :submit)
 
     get("/achievements", IncentivesController, :index_achievements)
+    get("/achievements/totalxp", IncentivesController, :calculate_totalxp)
```
angelsl, line 82: This probably makes more sense as `/user/total_xp` because the XP value includes assessment XP as well.

lib/cadet_web/controllers/incentives_controller.ex
```diff
@@ -2,20 +2,40 @@ defmodule CadetWeb.IncentivesController do
   use CadetWeb, :controller
 
   use PhoenixSwagger
+  import Cadet.Assessments
```
angelsl, line 5: Please don't import. Just `alias` it and use `Assessments.user_total_xp`

lib/cadet_web/controllers/incentives_controller.ex
```diff
@@ -2,20 +2,40 @@ defmodule CadetWeb.IncentivesController do
...
+  @spec index_goals(Plug.Conn.t(), any) :: Plug.Conn.t()
   def index_goals(conn, _) do
     render(conn, "index_goals_with_progress.json",
       goals: Goals.get_with_progress(conn.assigns.course_reg)
     )
   end
 
+  def calculate_totalxp(conn, _) do
+    course_id = conn.assigns.course_reg.course_id
+    total_achievement_xp = Achievements.get_total_xp(course_id)
```
angelsl, line 24: I'm quite confused. Is this supposed to be the total possible achievement XP or the total received achievement XP of the user?

On line 29, this is added to the user's total received assessment XP, so I presume you meant it to be the latter. But you only pass `course_id` to `get_total_xp`. That can't get you the total XP received by the user...

lib/cadet_web/controllers/incentives_controller.ex
```diff
@@ -2,20 +2,40 @@ defmodule CadetWeb.IncentivesController do
...
+    user = conn.assigns.current_user
+
+    if user.latest_viewed_course_id do
```
angelsl, line 27: You should not access `latest_viewed_course_id` here, because the route specifies a specific course ID.

What you probably want is `conn.assigns.course_reg.id`...

lib/cadet/incentives/achievements.ex
```diff
@@ -21,7 +21,43 @@ defmodule Cadet.Incentives.Achievements do
...
+  # To Do
+  Run elixir tests for edge cases etc.
+
+  This returns Achievement structs with prerequisites, goal association and progress maps pre-loaded.
+  """
+  def get_total_xp(course_id) when is_ecto_id(course_id) do
```
angelsl, line 31: This function is a bit strange. Is it supposed to return the total possible XP of all achievements in a course?

lib/cadet/incentives/achievements.ex
```diff
@@ -21,7 +21,43 @@ defmodule Cadet.Incentives.Achievements do
...
+
+    acc_goals = fn (all_goals) ->
+      Enum.reduce_while(all_goals, 1, is_all_goal_completed)
+    end
+
+    acc_achievements = fn (achievement_item, acc) ->
+      acc + acc_goals.(achievement_item.goals) * achievement_item.xp
+    end
+
+    Enum.reduce(achievements, 0, acc_achievements)
```
angelsl, line 57: Please don't declare helper functions like that, either use `defp` or indeed just inline the statements. These functions are small enough..

e.g.

```ex
    Enum.reduce(achievements, 0, fn (achievement_item, acc) ->

      completed_goals = Enum.reduce_while(achievement_item.goals, 1,
        fn (goal_item, acc) ->
          if Enum.count(goal_item.goal.progress) != 0 and Enum.at(goal_item.goal.progress, 0).completed,
            do: {:cont, acc},
            else: {:halt, 0}
        end)
      acc + completed_goals * achievement_item.xp
    end)
```

But this entire sequence is probably better written as:

```ex
    achievements
    |> Enum.map(fn ach -> 
        if Enum.all?(ach.goals, &(Enum.count(&1.goal.progress) != 0 and Enum.at(&1.goal.progress, 0).completed)),
          do: ach.xp,
          else: 0
      end)
    |> Enum.sum()
```

But then...

Your query at the top selects *all* goal progress entries in the course, for all users!

You need to pass in the course registration ID to this function, and adjust the query so that it only selects progress entries for the particular user you are calculating the XP for.

Finally.. I think this also doesn't handle achievements whose XP are dependent on the goal count.

angelsl (changes requested):

config/dev.exs
riccqi: Can we instead put this file into .gitignore?

angelsl: If you need to have different credentials, just override it in `dev.secrets.exs` in the exact same manner.

lib/cadet/incentives/achievements.ex
riccqi: Unintentional, changed it back

mix.lock
riccqi: What would be the best way to undo mix.lock changes?

lib/cadet/incentives/achievements.ex
riccqi: Most bugs here should be solved, still trying to comprehend the latter approach haha

mix.lock
angelsl:
```
git checkout master -- mix.lock
mix deps.get
```

lib/cadet/incentives/achievements.ex
angelsl: > still trying to comprehend the latter approach haha

In words:

1. For each achievement, take the achievement's XP award (`ach.xp`) if all goals are completed, else 0
2. Sum the numbers in the list resulting from step 1

angelsl (approved):

PR #906, opened 2022-09-07
add contest score job logging

YaleChen299:


martin-henz:
Mix credo failing. Can you take another look?

PR #914, opened 2022-10-13
✨ Feat: Add `all_users_total_xp` function and route

ShenyiCui:
## Changes
- Added a new route at `{{hostname}}/v2/courses/{{course}}/admin/users/total_xp` used to output all students total XP.


test/cadet_web/admin_controllers/admin_user_controller_test.exs
```diff
@@ -565,6 +565,743 @@ defmodule CadetWeb.AdminUserControllerTest do
     end
   end
 
+  describe "GET /v2/courses/{course_id}/user/total_xp" do
```
chownces, line 568: these are copied over from `user_controller_test.exs`? could we have some tests for the correct admin endpoint?

lib/cadet_web/admin_controllers/admin_user_controller.ex
```diff
@@ -289,4 +289,109 @@ defmodule CadetWeb.AdminUserController do
         end
     }
   end
+
+  def combined_user_total_xp(conn, _) do
+    course_id = conn.assigns.course_reg.course_id
+
+    combined_user_xp_total_query = """
```
chownces, line 296: could you place this query under the `Achievements` context instead of the controller layer? i.e. `lib/cadet/incentives/achievements.ex`

chownces (changes requested):

RichDom2185 (changes requested):

RichDom2185:
Related frontend PR (merged): https://github.com/source-academy/frontend/pull/2281

RichDom2185:
As discussed offline, we will close this PR and revert the already-merged frontend change.

PR #938, opened 2023-04-05
Removing unique score constraint for contest entries from backend (in preparation for frontend PR)

kjw142857:
Aim is to remove unique score checking for contest entries in backend in preparation for our frontend pull request to implement our contest voting tierlist, which accepts non-unique scores as submissions.

Am changing these 5 files:

assessments.ex - removed reference to non-unique submission votes causing an error and changed corresponding error message.
submission_votes_test.exs - removed test for unique votes.
submission_votes.ex - commented out unique score checker (unique_constraint)
Added new migration file 20230331010500_remove_unique_score_constraint.exs - reversible migration that removes the unique score index from the database, such that it can accept non-unique score entries.
answer_controller_test.exs - removed test case that checked for duplicate score, changed test case to check for updated error message in assessments.ex

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -703,8 +703,9 @@ defmodule Cadet.Assessments do
       {:error, :race_condition} ->
         {:error, {:internal_server_error, "Please try again later."}}
 
-      {:error, :vote_not_unique} ->
-        {:error, {:bad_request, "Invalid vote or vote is not unique! Vote is not saved."}}
+      # Description changed from :vote_not_unique and corresponding message changed.
+      {:error, :invalid_vote} ->
+        {:error, {:bad_request, "Invalid vote! Vote is not saved."}}
 
       _ ->
         {:error, {:bad_request, "Missing or invalid parameter(s)"}}
@@ -1012,7 +1013,7 @@ defmodule Cadet.Assessments do
   Computes rolling leaderboard for contest votes that are still open.
   """
   def update_rolling_contest_leaderboards do
-    # 115 = 2 hours - 5 minutes
+    # 115 = 2 hours - 5 minutes is default.
     if Log.log_execution("update_rolling_contest_leaderboards", Timex.Duration.from_minutes(115)) do
       Logger.info("Started update_rolling_contest_leaderboards")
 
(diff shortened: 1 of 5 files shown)


lib/cadet/assessments/assessments.ex
```diff
@@ -1630,7 +1631,8 @@ defmodule Cadet.Assessments do
     |> Repo.transaction()
     |> case do
       {:ok, _result} -> {:ok, nil}
-      {:error, _name, _changset, _error} -> {:error, :vote_not_unique}
+      # error type has been changed to :invalid_vote
```
YaleChen299, line 1634: I think we can remove this comment since we have reviewed this PR.

lib/cadet/assessments/assessments.ex
```diff
@@ -703,8 +703,9 @@ defmodule Cadet.Assessments do
       {:error, :race_condition} ->
         {:error, {:internal_server_error, "Please try again later."}}
 
-      {:error, :vote_not_unique} ->
-        {:error, {:bad_request, "Invalid vote or vote is not unique! Vote is not saved."}}
+      # Description changed from :vote_not_unique and corresponding message changed.
```
YaleChen299, line 706: It is Okay to remove the comment, since other developer might be buffled by the word "change", since they might not be following this project. This change history is captured in github so it is okay.

lib/cadet/assessments/submission_votes.ex
```diff
@@ -17,6 +17,7 @@ defmodule Cadet.Assessments.SubmissionVotes do
   @required_fields ~w(voter_id submission_id question_id)a
   @optional_fields ~w(score)a
 
+  # |> unique_constraint(:unique, name: :unique_score) (Unique constraint removed)
```
YaleChen299, line 20: For here, change the comment to a factual statement like "there is no unique constraint...blah blah".

test/cadet/assessments/submission_votes_test.exs
```diff
@@ -61,21 +61,6 @@ defmodule Cadet.Assessments.SubmissionVotesTest do
...
-
-      second_entry =
-        first_entry
-        |> Map.delete(:submission_id)
-        |> Map.put(:submission_id, new_submission.id)
-
-      {:error, changeset} = Repo.insert(second_entry)
-      refute changeset.valid?
-    end
+    # Removed test for unique submission vote
```
YaleChen299, line 64: similarly here

YaleChen299 (changes requested):
LTGM! Thank you for the good work! Just some minor nits for comments.

YaleChen299 (approved):

PR #943, opened 2023-06-02
Freeze GitHub runners' OS version

### LINKED ISSUE #942: Built releases will not run on AWS
As mentioned in [their blog](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/), `ubuntu-latest` workflows now runs on Ubuntu 22.04.

As we discovered only recently (as we have not deployed a new backend in a while), the GitHub-generated releases will not run on our servers. I suspect it's due to the above: Ubuntu 22.04 comes with `libc6-2.35` by default, so the GitHub runners will dynamically link those new versions of those libraries into our compiled release.

This is a problem as our AWS hosts currently run Ubuntu 20.04, which come packaged with `libc6-2.31`. Thus, trying to deploy our backend results in the following error (truncated):

```
...version `GLIBC_2.34' not found...
```

RichDom2185:
See #942 for more info.

Let us freeze the OS versions to Ubuntu 20.04 to keep them consistent with what we are currently using on our AWS servers.

Fixes #942.


YaleChen299 (approved):
LGTM

Issue #945, opened 2023-06-04
Upgrade to Erlang/OTP 25 or 26


RichDom2185:
With Erlang/OTP 26 released earlier this year in [May](https://www.erlang.org/news/164), we are now 2 (prod) or 3 (dev) major versions behind.

Let us try to update the codebase to newer versions of Erlang/OTP.

PR #946, opened 2023-06-04
Bump Elixir and Erlang versions

### LINKED ISSUE #942: Built releases will not run on AWS

### LINKED ISSUE #944: Inconsistency between CI/CD and README versions
In our CI/CD pipelines, we see the following:
* https://github.com/source-academy/backend/blob/15e12d80f945b225cf88b389958a260285c879f1/.github/workflows/ci.yml#L26-L27
* https://github.com/source-academy/backend/blob/15e12d80f945b225cf88b389958a260285c879f1/.github/workflows/ci.yml#L30

But in our README for development:

* https://github.com/source-academy/backend/blob/15e12d80f945b225cf88b389958a260285c879f1/README.md#L17-L19

Note that the README shows Erlang/OTP 23 while the actual build uses Erlang/OTP 24. Judging by the commit/PR history, it seems like even during the upgrade to OTP 24 (#856), the inconsistency was already there to begin with.

Let us make these versions consistent to potentially catch compatibility bugs during development.

RichDom2185:
I am keeping #945 open to investigate the possibility to upgrade to Erlang/OTP 26. This is merely a hotfix to resolve #942.

Also resolves #944.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -24,8 +24,8 @@ jobs:
     env:
       MIX_ENV: prod
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      ELIXIR_VERSION: "1.13.3"
-      OTP_VERSION: "24.2.1"
+      ELIXIR_VERSION: 1.13.4
+      OTP_VERSION: 25.3.2
     steps:
       - uses: rlespinasse/github-slug-action@v3.x
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


YaleChen299 (approved):

RichDom2185:
Workflow compiles so should be fine to merge.

PR #950, opened 2023-06-14
Add Endpoints for Notifications

Santosh3007:
## Changes

Adding extra endpoints needed for notifications


lib/cadet_web/router.ex
```diff
@@ -169,6 +169,19 @@ defmodule CadetWeb.Router do
     )
   end
 
+  # Notifications endpoints
```
YaleChen299, line 172: How are these APIs being called from the frontend? Why are they not piped through authentication and course-specific assignment? I am not sure what might happen here. If some request is fired from a student that already has his log in session expired or if the student changed course(which might lead to some cache issue or privilege issue)...

YaleChen299: Also please separate the admin routes from the student routes

lib/cadet_web/controllers/new_notifications_controller.ex
```diff
@@ -0,0 +1,116 @@
...
+
+  def delete_time_options(conn, params) do
+    # time_option = Repo.get(TimeOption, time_option_id)
+
+    # if is_nil(time_option) do
+    #   conn |> put_status(404) |> text("Time option of given ID not found")
+    # end
+
+    # case Repo.delete(time_option) do
+    # JUNYI AND SANTOSH: NOT DEFINED???
```
YaleChen299, line 104: please clean up unused comments as well.

YaleChen299 (changes requested):
Thanks for the quick work! However, I think the design needs major changes. 

First of all, try to follow the current routing scopes, if a notification belongs to a course, semantically it should be under /course. In that way, you will naturally pipe through authentication checks, course checks sand staff checks. 

Second, please separate admin routes from student routes (or do you want to separate them from staff also?). Do think about the privilege of each route. The backend needs to be intact without the assumption of the well-behaviour of the frontend.

Lastly, please clean up unnecessary comments and use query styles in the piping fashion(which aligns with the current code base). Try to use Logger instead of IO for logging.

YaleChen299:
This PR is superseded by #973. Hence, this will be closed when #973 is merged.

RichDom2185:
Closing as per offline discussion.

PR #965, opened 2023-07-07
Add Publish Grading Feature

emptygx:
**To be merged with Frontend [#2856](https://github.com/source-academy/frontend/pull/2856)**

For implementation decisions, see Frontend PR.

## Changes
- Assessment Table: Add new is_grading_published field which determines if students can see their grades.
- Assessment Config Table: Add new is_grading_auto_published field which determines if grades are auto published for auto graded assessments.
- Backend changes xp/grading details, e.g. grader, autograding_status, back to default if is_grading_published is false. This prevents students from checking response from server to see their grades even though the Frontend hides it. 
- Notifications Types: Removed autograded and graded types. Added published and unpublished grading types.
- Notifications now created when grades are published/unpublished
- Implement relevent functions for publishing/unpublishing grades
- Admins are able to publish/unpublish all graded submissions
- Autograder can autopublish grades when all answers have a `success` autograding_status
- Updated seed to accommodate changes


YaleChen299:
> > Hi! Please fix the format and CI and tag me again when the PR is ready. Thank you!
> 
> @YaleChen299 Hello! Sorry, I have fixed the formatting. Thank you!

No worries! Can you check the CI again? I will approve the CI run when I see then.

YaleChen299 (dismissed):
Hi @emptygx, I have fixed the formatting and credo issues. However, a few tests are not passing. I suspect that the behaviour of `user_total_xp` is affected by the behaviour of published/unpublished grades. 

What is the expected behaviour for the published/unpublished grades? Is it that users will not see their grades if unpublished? 

Could you look into these test cases and fix them? Please document the correct behaviours in terms of the test cases, i.e. for any behaviours that may be affected by the publish grading feature, create at least one positive and one negative test case. 

Thank you so much for wrapping up this properly! I think this feature would make the grading activities much more user-friendly!

lib/cadet/accounts/notification_type.ex
```diff
@@ -10,5 +10,7 @@ defenum(Cadet.Accounts.NotificationType, :notification_type, [
   # Notifications for unsubmitted submissions
   :unsubmitted,
   # Notifications for submitted assessments
-  :submitted
+  :submitted,
+  # Notifications for unpublished grades
+  :unpublishedGrading
```
YaleChen299, line 15: let's keep the atom as snake case, as camel case is used for module names which are under the hood atoms.

lib/cadet/assessments/assessments.ex
```diff
@@ -248,6 +250,14 @@ defmodule Cadet.Assessments do
         Answer
         |> join(:inner, [a], s in assoc(a, :submission))
         |> where([_, s], s.student_id == ^course_reg.id)
+        # change a.xp and a.xp_adjustment to 0 if a.is_grading_published is false
+        # and return all answers
+        |> select([a, s], %{
+          a
+          | xp: fragment("CASE WHEN ? THEN ? ELSE 0 END", s.is_grading_published, a.xp),
```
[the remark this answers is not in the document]
YaleChen299: another thing you can do is just query the answers where the `a.is_grading_published` is true since we are only aggregating here.

lib/cadet/assessments/assessments.ex
```diff
@@ -844,6 +862,114 @@ defmodule Cadet.Assessments do
...
+    # allows staff to unpublish own assessment
+    bypass = role in @bypass_closed_roles and submission.student_id == course_reg_id
+
+    with {:submission_found?, true} <- {:submission_found?, is_map(submission)},
+         {:allowed_to_unpublish?, true} <-
+           {:allowed_to_unpublish?,
+            role == :admin or bypass or
+              Cadet.Accounts.Query.avenger_of?(cr, submission.student_id)} do
+      Multi.new()
+      |> Multi.run(
```
YaleChen299, line 884: I think there is no need to use multi and transaction here since there is only a single update. May II ask what was the reason for this?

lib/cadet_web/admin_controllers/admin_assets_controller.ex
```diff
@@ -22,7 +22,7 @@ defmodule CadetWeb.AdminAssetsController do
 
     case Assets.delete_object(Courses.assets_prefix(course_reg.course), foldername, filename) do
       {:error, {status, message}} -> conn |> put_status(status) |> text(message)
-      _ -> conn |> put_status(204) |> text('')
+      _ -> conn |> put_status(204) |> text(~c"")
```
YaleChen299, line 25: Hi, what is the reason for this change? This should be unrelated to this PR?

lib/cadet/devices/devices.ex
```diff
@@ -208,7 +208,7 @@ defmodule Cadet.Devices do
               },
               300,
               [],
-              ''
+              ~c""
```
YaleChen299, line 211: similarly here, this is unrelated to this PR right? Is it cuz it is failing for some test?

emptygx: Hi, yes it was failing the formatting checks and was added when I ran mix format

lib/cadet_web/admin_controllers/admin_assets_controller.ex
emptygx: Same as the other one it was auto-corrected while formatting

lib/cadet_web/admin_controllers/admin_grading_controller.ex
```diff
@@ -82,6 +82,34 @@ defmodule CadetWeb.AdminGradingController do
     end
   end
 
+  def unpublish_grades(conn, %{"submissionid" => submission_id}) when is_ecto_id(submission_id) do
```
YaleChen299, line 85: hi please add tests for the controllers thanks!

YaleChen299:
> (part of frontend issue #2522)
> 
> Currently, the grading states design and business logic cause ambiguity with resubmission of assessments being assumed to be graded immediately, and with grades being displayed when they should not.
> 
> This feature aims to solve this issue by introducing a new grading state `published` and a new backend boolean `is_grading_published`, allowing explicit publishing of student's grades on the frontend and preventing the fetching of grades if they have yet to be published, hence creating a better logic flow.
> 
> This pull request contains changes to be made to the backend to enable this new feature.

Re-thinking about the problem that we are trying to solve, I think we need to auto un-publish the grades when we unsubmit.
The main point of publish grading is to solve resubmitted submissions shown as `graded` to graders so that grader will not know if the student is graded after resubmission.
This logic is a bit different then whether the grade is publish to the student or not. Please correct me if i am wrong! Thanks!

emptygx:
> Re-thinking about the problem that we are trying to solve, I think we need to auto un-publish the grades when we unsubmit. The main point of publish grading is to solve resubmitted submissions shown as `graded` to graders so that grader will not know if the student is graded after resubmission. This logic is a bit different then whether the grade is publish to the student or not. Please correct me if i am wrong! Thanks!

Yes, I agree that should be part of the feature, which has already been implemented as part of the logic on the front end, which auto-unpublishes grades whenever an assessment is unsubmitted. Sorry for not mentioning that in the PR description, will make sure to add it into future documentation! Should I add it to the PR description as well?

lib/cadet/assessments/assessments.ex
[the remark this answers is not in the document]
emptygx: I found this to be the easier and more straightforward way to filter the data, but if maintainability may be an issue, then doing it at the controller level might be better like you said. I am not too sure how to go about that though and may need some guidance!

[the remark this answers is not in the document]
YaleChen299: I agree that making this filtering at the controller level is a bit more troublesome. I think what we can do is at least separate this filtering from the main subquery selection. If you can make it into a function to be piped through with sematic name like `filter_published_xp`, I will be happy to accept that.

GabrielCWT:
TODO:

Implement test for is_grading_published after this and #1088 are merged

test/cadet_web/admin_controllers/admin_grading_controller_test.exs
josh1248, line 1: what a legend, frontend definitely needs someone who cares about tests 🌚

lib/cadet/assessments/assessments.ex
```diff
@@ -1094,11 +1155,272 @@ defmodule Cadet.Assessments do
...
+      Enum.each(submissions, fn submission ->
+        Notifications.handle_unpublish_grades_notifications(
+          assessment_id,
+          Repo.get(CourseRegistration, submission.student_id)
+        )
+      end)
+
+      {:ok, nil}
+    else
+      {:error, {:forbidden, "Only Admin is permitted to publish all grades"}}
```
RichDom2185, line 1420: Typo: Incorrecr error message for unpublish all.

lib/cadet/assessments/assessments.ex
```diff
@@ -1094,11 +1155,272 @@ defmodule Cadet.Assessments do
...
+        })
+
+      question_query =
+        Question
+        |> group_by([q], q.assessment_id)
+        |> join(:inner, [q], a in Assessment, on: q.assessment_id == a.id)
+        |> select([q, a], %{
+          assessment_id: q.assessment_id,
+          question_count: count(q.id)
+        })
```
RichDom2185, line 1350: Correct me if I'm wrong, but since these are subqueries, they'll run first in entirety before any downstream filtering.

So in order to avoid the ~30s query from the joining of the large table, should we filter by course ID here first?

RichDom2185 (commented):
Hi, thanks for working on this! Just some comments below, thanks!

lib/cadet/assessments/assessments.ex
GabrielCWT: I managed to filter the Question table subquery but the Answer table subquery requires quite a few joins to obtain the course_id to filter by. I was wondering if it is worth the multiple joins since I'm not too sure about costs in SQL.

Joins:
Answer -> Question -> Assessment (course_id)

I also thought of combining the queries together since I already use Question -> Assessment but you cant use aggregate functions e.g. filter and count

RichDom2185 (approved):
LGTM! Let's merge this first to not block the other team's PRs. As to your question, let's discuss it after this is merged.

Thanks for the work!

PR #970, opened 2023-07-11
Include username and provider in JWT token

RichDom2185:
Previously, the JWT claims only include the user ID (the primary key as defined in the main backend's DB).

However, as we move to a microservices-based architecture, we would like to make the main backend a centralised sole issuer of JWT tokens that can be independently validated by each service across the entire cluster, and thus we need to make this token agnostic/independent of the DB/issuer.

Part of https://github.com/source-academy/stories-backend/issues/16.

@YaleChen299 to discuss: Should we just change the existing main backend behaviour to simply rely on the username and provider instead of the primary key?


RichDom2185:
Once this is deployed, all existing users would get 400 Bad Request and have to log out and log back in.

YaleChen299 (approved):
Thanks! This looks good to me! As for removing user_id, I don't see much incentive to do so. This will likely not cause future problem as well, and is easy to remove in the future when we feel the need to.

PR #971, opened 2023-07-13
Update CD Workflow to archive previous releases

### LINKED ISSUE #941: Archive previous deployment releases
Currently, the GitHub Actions workflow will delete all the past releases and only retain the most recent release:

image

This is problematic as it makes it hard to rollback to a specific version without manually looking through the GitHub Actions history and force-pushing the branch back to the exact older state. Even then, this solution would not be ideal as it is essentially a "hacky" way to make GitHub Actions rebuild the particular release. If environment settings (e.g. runner OS version, environment secrets/variables) have changed since the last time that particular release was built, the resulting compiled release may not be identical.

We should change this behavior.

RichDom2185:
Closes #941.

Changes:

* Use `github.ref_name` instead of unnecessary `rlespinasse/github-slug-action`
* Auto-generate release tags and titles based on the current time
* Generate changelog from the last release from the same branch
* Archive previous releases using the autogenerated tag name

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -16,6 +16,8 @@ on:
       - "mix.lock"
       - ".github/workflows/cd.yml"
       - "!config/*.example"
+env:
+  TIME: ${{ github.event.repository.pushed_at }}
 
 jobs:
   ci:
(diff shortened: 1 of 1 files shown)


RichDom2185:
@YaleChen299 the issue now is:

* The releases are tagged with the time the workflow was run, not `latest-[branch_name]`
* While the `latest-[branch_name]` tag exists and points to the same commit, it does not point to the release,
* So the uploaded asset can't be found when querying from `latest-[branch_name]`

PR #973, opened 2023-07-18
Fix some router security and code style issues

Catherine9898:
1. Follow the current routing scopes, the notification belongs to a course, they will naturally pipe through authentication checks, course checks sand staff checks.
2. Separate admin routes from student/staff routes. 
3. Clean up unnecessary comments and use query styles in the piping fashion (which aligns with the current code base). 
4. Try to use Logger instead of IO for logging.


RichDom2185:
Closing as per offline discussion.

PR #979, opened 2023-07-30
Llm default prompt

clementlemon02:
Adding LLM default prompt column to xml files


RichDom2185 (commented):
Can you run `mix format` to fix the formatting?

PR #984, opened 2023-08-13
Llm default branch

clementlemon02:
add default prompt to assessments


lib/cadet/devices/devices.ex
RichDom2185, line 1: Why is this file deleted?

test/support/xml_generator.ex
RichDom2185, line 1: This is just a test case, is there no corresponding update of the actual XML parser?

RichDom2185 (changes requested):
This PR seems a little weird, is it supposed to contain anything else?

It seems to only contain style changes to existing code (as well as a deletion of `devices.ex`, which is not only unrelated to LLM, but also critical).

I'll close this for now; feel free to reopen should this change.

PR #986, opened 2023-08-23
Update grading json to include student usernames

chownces:
As per title. To facilitate the display of student usernames in the grading table and GradingWorkspace in the frontend.


RichDom2185 (approved):
LGTM, thanks!

PR #991, opened 2023-09-03
Fix swagger paths

RichDom2185:
Uses `router.exs` as the source of truth and fixes any inconsistencies between those routes and the swagger path entries.

If I have missed any, they will be done in a future PR as this is a good-to-have and non-critical.

There are also some invalid names that have been changed to point to the correct ones.


YaleChen299 (approved):
Thanks for the fix!

PR #1000, opened 2023-09-14
Fix issue with contest voting pooling

kjw142857:
PR to fix issue with the voting file only taking the entries submitted before point of upload and not until the contest closes.

One of these two "trigger events" will possibly release entries into the voting pool:
- During upload of the voting file, the contest is checked to see if it is still open. If it is closed, immediate release of entries into the voting pool. If it is still open, release of entries into the voting pool is put on hold.
- Every day at 00:01, batch job runs to check if any contest voting has not done the entry assignment yet. If there is, checks whether the corresponding contest has closed and if it has closed already, releases entries into the voting pool.


RichDom2185 (commented):
Overall looks okay, but perhaps instead simply just modifying the existing test cases to account for this behaviour, could you create new test cases to explicitly account for this new change? The test coverage is dropping by 0.5% otherwise, which is quite significant.

Thanks!

kjw142857:
Update: have added the tests! Coverage is now -0.05%

lib/cadet/assessments/assessments.ex
```diff
@@ -500,6 +501,35 @@ defmodule Cadet.Assessments do
...
 
+  def update_final_contest_entries do
+    # 1435 = 1 day - 5 minutes
+    if Log.log_execution("update_final_contest_entries", Duration.from_minutes(1435)) do
+      Logger.info("Started update of contest entry pools")
+      questions = fetch_unassigned_voting_questions()
+
+      for q <- questions do
+        insert_voting(q.course_id, q.question["contest_number"], q.question_id)
+      end
```
RichDom2185, line 512: Question: Should the return value of this function be checked here? What happens if this line throws an error?

E.g. what if when the voting entry was originally inserted, the number of the contest/assessment in question still exists, but is sometime later deleted? This line would fail as a result, right?

RichDom2185 (approved):
LGTM. This does mean that if a contest closes in the middle of the day, voting for that contest cannot commence until at least the day after (unless the voting file is only uploaded after the assessment itself closes, then the voting will then begin immediately).

Thanks for the fix and the comprehensive tests!

lib/cadet/assessments/assessments.ex
RichDom2185: As discussed offline, deleting the assessment will delete the question, so the for loop would be empty, so no error will be thrown.

Thanks!

RichDom2185: Update: Deleting the assessment will delete the assessment question, but not the voting question (which is the one we are interested in!). #1002 will resolve this by enforcing a precondition for this loop.

PR #1002, opened 2023-09-17
Prevent deletion of contests when contest voting file is present

kjw142857:
Modified delete_assessment so that if a corresponding contest voting file has been uploaded, deleting the contest file is not allowed (400 error code) until the contest voting file has been deleted first.

In addition, added an additional check during retrieval of fetch_unassigned_voting_questions that each question must reference an existing contest.


RichDom2185 (commented):
Looks good, but can we add tests, please? Thanks a lot!

kjw142857:
Update:

- Tests have been added
- Additional check during update_final_contest_entries has been added to ensure that insert_voting is never called if the voting question references a nonexistent contest, which could possibly happen due to data corruption or direct modification of the DB

RichDom2185 (approved):
LGTM, thanks!

PR #1004, opened 2023-09-20
Fix tests failing on some random seeds

kjw142857:
Modify assertion in #1002 - testing for fetching of valid contest questions to check for permutations instead of a direct ==.


RichDom2185 (approved):
LGTM, thanks for working on this so quickly!

PR #1010, opened 2023-10-01
Change modifier to 80 for Game of Tones contest

kjw142857:
Change the token penalty modifier from 50 to 80 for Contest C6 Game of Tones, NUS SA 23/24 Sem 1 at request of educators.

Note: In the future, proposed implementation is to have the token penalty modifier inserted in the contest voting file as a parameter at the point of upload.


RichDom2185 (changes requested):
I believe the denominator should be 80, not 70, as per the announcement given on Canvas.

RichDom2185 (approved):
LGTM

PR #932, opened 2023-03-07
Add Avenger Backlog Notifications

Santosh3007:
# Source Academy Notifications

This is the first iteration of introducing notifications into Source Academy. Currently, this introduces two notification types: `Avenger Backlog` , `Assessment Submission` and only through the email medium.

In future iterations, more notifications can be included and other mediums of notifications can be explored.

## Changes

- Implement Notifications System Framework
- Add Oban (Job Processing Library)
- Add Bamboo (Email Dispatch)
- Implementation for avenger backlog notifications
- Implementation for assessment submission notifications
- Tests for Bamboo, Oban & notification models

## Notes
* The original `notifications` table is left untouched for now
* 2 triggers (and procedures) are added to the database to populate `notification_configs` when new courses are insert & when new assessments are created
* Migration added to populate `email` column in `users` table using `username` (Only for NUS users where the provider is luminus)
* Will improve robustness of Oban/Mailer testing in future PRs

PR #1031, opened 2023-11-06
Fix DB crash due to timeout

RichDom2185:
Our grading summary uses a long, raw SQL query to generate the JSON directly, bypassing our ORM. This causes issues when we have hundreds of thousands of answers to query.

Lots of details are glossed over in this PR as they have already been discussed internally.

Changes:

* Create DB index – improves read performance
* Splits single SQL query into three – more generous time limit/less likely to time out
* (Mostly) decouples M, V, C layers for grading summary
  * Create separate view layer for admin grading summary view
  * Removes unnecessary joins to dynamically generating based on view model – also reduces size of data to be passed from DB to backend layer
* Migrates JSON view generation from DB layer to backend (view) layer – reduces DB total query time

While request-response delay is only slightly improved, DB query time is improved much more significantly, as a lot of the view generation logic is done in the backend layer now. The DB only provides the bare minimum amount of data necessary to generate this view.

Tested and JSON output is of the exact same structure compared to the old query.

---

***CAVEAT***

This PR will break Avenger backlog emails. Not just because I removed the "ungraded only" functionality (which was only added in #932 and used in emails), but while compile errors are fixed to match the new function signature, it will most definitely result in a runtime error (as the email templates are not updated to reflect the new structure.

The reason I think it's fine is:

* Email notifications are not used anywhere yet
* I really think #932 is the wrong approach to generate the emails:

  image

  `all_submissions_by_grader_for_index` generates a view (back then was still the JSON string), which is then converted back to a "model" using `Jason.decode`, which is then further manipulated, and the final view is generated using the email template.

  My proposal would be to create a separate function entirely for this functionality, which would optimise everything as well. But that is out of scope of this, and will come in a follow-up PR instead.


lib/cadet/assessments/assessments.ex
```diff
@@ -1238,114 +1238,106 @@ defmodule Cadet.Assessments do
...
+             ) #{group_filter};
+           """) do
+        {:ok, %{columns: columns, rows: result}} ->
+          result
+          |> Enum.map(
+            &(columns
+              |> Enum.map(fn c -> String.to_atom(c) end)
+              |> Enum.zip(&1)
+              |> Enum.into(%{}))
+          )
```
YaleChen299, line 1311: do we put a default case here? in case of error

YaleChen299 (approved):
Thanks for the fix!

lib/cadet/assessments/assessments.ex
RichDom2185: I didn't think it was necessary as I don't think the SQL will ever error (unless the DB server itself is down, in which case it would give the same error messages as the ones we have been having recently).

What do you think?

PR #1034, opened 2023-11-08
Add token divider field to contest voting XML file

kjw142857:
This pull request aims to allow educators to specify the divider for the token penalty applied when contest entries are scored. The divider will be specified in the contest XML file under the token_divider field, which is compulsory and will be parsed by the XML parser before later being referenced when calculating the formula score. If the token_divider field is not specified or is <= 0, the BE will not allow the XML contest voting file to be uploaded.

Note: The _smaller_ the token divider, the larger the penalty that will be applied for long programs. For most purposes, setting the token divider in the range 50-100 should be sufficient.


RichDom2185 (approved):
Sorry for the late review, LGTM!

Thanks a lot, this will definitely be very helpful to minimise redeployments when the denominator needs to change.

PR #1053, opened 2024-01-16
Set whether stories is enabled in course config

RichDom2185:
A backwards-compatible change thus can be merged independently of the frontend (associated frontend PR TBC).

This aims to allow the frontend, which first retrieves the course config from the main backend, to also check if it should then ping the stories backend.

This would solve the spurious 404 errors when stories is not enabled for a particular course.


RichDom2185 (commented):
@YaleChen299 since the config change is a relatively passive thing, I don't see the need to add more tests when the views are effectively identical. Can I get your thoughts on this? Thanks!

YaleChen299 (approved):
LGTM! Thank you so much!

PR #1072, opened 2024-02-24
Fix bug causing build_user_filter to filter incorrectly.

GabrielCWT:
- Changed build_user_filter to use correct table. student_id in Submissions refers to the Course Reg ID, not the User ID.


RichDom2185 (approved):
LGTM, thanks!

PR #1127, opened 2024-05-19
docs: Update schema in README

GabrielCWT:
This PR updates the schema.png in the README to reflect new tables and columns, e.g. teams and team_members tables


RichDom2185 (approved):
Thanks a lot, I think it makes sense to remove the constraints and the types from the diagram as we are interested in mainly the relationships between the various tables as opposed to the internal implementation; one can always just look at the DB to find out what the enforced constraints/types are.

LGTM!

PR #1144, opened 2024-08-04
Add LLM improvements

RichDom2185:
- [ ] Access control
- [ ] Logging & auditing
- [x] Move prompt generation to BE to prevent bypass


RichDom2185:
@yiwen101 the prompt builder is separated from SICP notes so that we can put AG logic inside the builder and retain the abstraction

lib/cadet_web/plug/rate_limiter.ex
```diff
@@ -0,0 +1,24 @@
+defmodule CadetWeb.Plugs.RateLimiter do
+  import Plug.Conn
+
+  @rate_limit 500
```
RichDom2185, line 5: Does this mean 500 requests every 24 hours?

yiwen101 (approved):
LGTM

RichDom2185:
Corresponding FE PR: https://github.com/source-academy/frontend/pull/3010

PR #1145, opened 2024-08-04
llm rate limiting

yiwen101:
Add rate limiting to the gpt chat api endpoint.
The current decision of "limiting to 500 calls per day per user" is arbitrary and open to discussion.


RichDom2185 (approved):
LGTM, thanks a lot!

PR #1151, opened 2024-08-12
Add payload size limit

### LINKED ISSUE #1150: Limit LLM individual message payload size
While the context size is limited to 20 messages in #1144, individual messages still can be arbitrarily long.

The controller should check for message length and limit to e.g. 1000 characters maximum, return a `422 Unprocessable Entity` status code otherwise.

yiwen101:
Closes #1150


lib/cadet_web/controllers/chat_controller.ex
```diff
@@ -51,45 +53,54 @@ defmodule CadetWeb.ChatController do
     response(200, "OK")
     response(400, "Missing or invalid parameter(s)")
     response(401, "Unauthorized")
+    response(421, "Message exceeds the maximum allowed length")
```
RichDom2185, line 56: Should this be 422 instead?

test/cadet_web/controllers/chat_controller_test.exs
```diff
@@ -76,6 +76,24 @@ defmodule CadetWeb.ChatControllerTest do
       end
     end
 
+    @tag authenticate: :student
+    @tag requires_setup: true
+    test "The content length is too long",
+         %{conn: conn, conversation_id: conversation_id} do
+      assert conversation_id != nil
+      max_message_length = ChatController.max_content_length()
+      message_exceed_length = String.duplicate("a", max_message_length + 1)
```
RichDom2185, line 85: Let's also add another testcase to test that `String.duplicate("a", max_message_length)` passes normally

lib/cadet_web/views/chat_view.ex
```diff
@@ -1,8 +1,12 @@
 defmodule CadetWeb.ChatView do
   use CadetWeb, :view
 
-  def render("conversation_init.json", %{conversation_id: id, last_message: last}) do
-    %{conversationId: id, response: last}
+  def render("conversation_init.json", %{
+        conversation_id: id,
+        last_message: last,
+        max_content_size: size
```
RichDom2185, line 7: I'm thinking if there is a need to configure it this way (as opposed to hardcoding), but I am fine with this as well. What do you think?

lib/cadet_web/controllers/chat_controller.ex
RichDom2185, line 1: Minor nit: Can we combine the `if` into the `with`? See https://elixirforum.com/t/proper-way-of-handling-simple-booleans-in-with-statement/45556/4 for an example

yiwen101: Thank you for pointing me to this. Fixed

test/cadet_web/controllers/chat_controller_test.exs
yiwen101: Added in later commit. However, in current setup the expected result is similar to the "conversation id does not match user id" case. Do you know how to fix this?

RichDom2185 (approved):
LGTM, thanks!

PR #1155, opened 2024-08-19
fix sorting for grading

thortol:
### Description

Currently there is a bug in source academy grading table where users are unable to sort by username and group

### Changes
- There was a bug where a left join condition was done on `user.id == submissions.student_id` and `user.id == courseRegistration.user_id`
- However, in the schema the conditions are `courseRegistration.id == submissions.student_id` and `courseRegistration.user_id == user.id`
- An additional bug was found during testing where the same row kept appearing every page. This happens when `submission.inserted_at` is the same which prevents LIMIT from working as the sort may be inconsistent.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### Checklist

- [x] I have tested this code

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -1943,12 +1943,12 @@ defmodule Cadet.Assessments do
         left_join: asst in subquery(question_answers_query),
         on: asst.assessment_id == s.assessment_id,
         as: :asst,
-        left_join: user in User,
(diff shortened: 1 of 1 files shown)


RichDom2185 (approved):
LGTM, thanks!

PR #1156, opened 2024-08-21
feat: Include hoursBeforeEarlyXpDecay in response

GabrielCWT:
Includes hoursBeforeEarlyXpDecay in return response. This is to allow the frontend to know when full bonus xp ends. 
Frontend PR: https://github.com/source-academy/frontend/pull/3025


lib/cadet_web/views/assessments_view.ex
```diff
@@ -37,7 +37,8 @@ defmodule CadetWeb.AssessmentsView do
       maxTeamSize: :max_team_size,
       hasVotingFeatures: :has_voting_features,
       hasTokenCounter: :has_token_counter,
-      isVotingPublished: &is_voting_assigned(&1.id)
+      isVotingPublished: &is_voting_assigned(&1.id),
+      hoursBeforeEarlyXpDecay: & &1.config.hours_before_early_xp_decay
```
RichDom2185, line 41: Is the extra `&` here intentional?

GabrielCWT: > Is the extra `&` here intentional?

Yep. It is an anonymous function for the `transform_map_for_view` function being used. This is because Map.get cannot access nested maps.

RichDom2185 (approved):
LGTM, thanks!

PR #1157, opened 2024-08-22
fix: Auto publish for non-autograded assessments

GabrielCWT:
Original implementation only auto publishes for autograded assessments. When updating the questions during grading, we now check if all questions are graded. If they are and auto publish config is enabled, publish the submission


RichDom2185 (approved):
LGTM, thanks a lot!

PR #1158, opened 2024-08-23
fix: Include permission check for auto publishing grades

GabrielCWT:
Auto publishing should follow the same workflow as the publish feature and not bypass any checks.


RichDom2185 (approved):
LGTM, thanks!

PR #1163, opened 2024-08-26
fix: Fix notifications of type "submitted" not being read

GabrielCWT:
Currently the notifications are not marked as read for type: submitted. When the grading is published, be it manually or automatically, the notification will be marked as read. The grading notification should also be marked as read when the submission is unsubmitted.

I feel that notifications should signify that there are things to be done on the grading page. This solution resolves this bug of there being "things to do" even though there aren't any.
Screenshot 2024-08-26 at 10 31 20 PM

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -1139,6 +1139,13 @@ defmodule Cadet.Assessments do
           )
       end
 
+      # Remove grading notifications for submissions
+      Notification
+      |> where(submission_id: ^submission_id, type: :submitted)
+      |> select([n], n.id)
+      |> Repo.all()
+      |> Notifications.acknowledge(cr)
+
       {:ok, nil}
     else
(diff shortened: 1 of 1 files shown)


GabrielCWT:
One thing I have noticed today: Paths also create a notification. Would it be a good idea to add a check where if `is_manually_graded` is true then create a notification, else do not. This is to follow the idea that the todo count is meant to be for things the staff needs to do in the grading tab.

RichDom2185:
> One thing I have noticed today: Paths also create a notification. Would it be a good idea to add a check where if `is_manually_graded` is true then create a notification, else do not. This is to follow the idea that the todo count is meant to be for things the staff needs to do in the grading tab.
> 
> Furthermore, there is no way to resolve the todo for a path since it is autograded and autopublished

This can be a separate issue + PR, thanks!

RichDom2185 (approved):
LGTM, thanks!

PR #1165, opened 2024-08-29
fix: Fix unintended auto-publishing of autograder

GabrielCWT:
Currently the autograder autopublishes when all answers are successfully autograded and if `is_grading_auto_published` is true.

Autograder should only be auto publishing when there is no need for manual intervention as well. Therefore, we should also be checking if `is_manually_graded` is false.

Closes: https://github.com/source-academy/backend/issues/1164


RichDom2185 (approved):
LGTM, thanks a lot!

PR #1167, opened 2024-08-31
feat: Add check to notification creation

GabrielCWT:
This PR adds a check to finalise submission. It will only create a notification if the assessment requires manual grading. This would mean that the staff has a TODO and thus a notification should be created.

In the context of CS1101S: Paths are autograded and require no manual grading. Without this check, staff will receive a notification under the grading tab but there is no action required. There is also no way to resolve the notification easily as the submission is autograded and autopublished. 

Closes: https://github.com/source-academy/backend/issues/1166


GabrielCWT:
I am willing to discuss alternative solutions if this does not seem like the correct approach to the issue.

RichDom2185 (approved):
LGTM, we can test this and take a look at the user feedback, thanks!

Issue #1170, opened 2024-08-31
Wrong value passed to update bonux XP


RichDom2185:
<img width="769" alt="image" src="https://github.com/user-attachments/assets/1f3a7890-3515-49c7-a9c3-b07411d999f5">

Should be passing the updated submission. Otherwise submitted at is just `nil` all the time.

PR #1172, opened 2024-08-31
Fix issues with bonus xp

### LINKED ISSUE #1168: Unnecessary logic
image

Grouping by answer ID (the PK) gives one result, so no need to aggregate, can just select XP etc. directly?

thortol:
### Description

Resolves #1168, #1170.

Some issues with the code and fixes the bug where autograded assignments do not have bonus xp awarded. The bug is fixed by updating bonus xp when grading is published.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### Checklist

- [x] I have tested this code

### DIFF
--- a/lib/cadet/assessments/assessments.ex
+++ b/lib/cadet/assessments/assessments.ex
@@ -1455,11 +1455,10 @@ defmodule Cadet.Assessments do
         Answer
         |> where(submission_id: ^submission_id)
         |> order_by(:question_id)
-        |> group_by([a], a.id)
         |> select([a], %{
           # grouping by submission, so s.xp_bonus will be the same, but we need an
           # aggregate function
-          total_xp: sum(a.xp) + sum(a.xp_adjustment)
+          total_xp: a.xp + a.xp_adjustment
         })
 
       total =
@@ -1470,8 +1469,6 @@ defmodule Cadet.Assessments do
         })
         |> Repo.one()
 
-      xp = decimal_to_integer(total.total_xp)
-
       cur_time =
         if submission.submitted_at == nil do
           Timex.now()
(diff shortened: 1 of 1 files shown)


lib/cadet/assessments/assessments.ex
```diff
@@ -1455,11 +1457,10 @@ defmodule Cadet.Assessments do
         Answer
         |> where(submission_id: ^submission_id)
         |> order_by(:question_id)
-        |> group_by([a], a.id)
         |> select([a], %{
           # grouping by submission, so s.xp_bonus will be the same, but we need an
```
RichDom2185, line 1461: This comment doesn't make sense, seems lifted from another code segment?

lib/cadet/assessments/assessments.ex
```diff
@@ -1470,15 +1471,20 @@ defmodule Cadet.Assessments do
...
         else
           submission.submitted_at
         end
 
+      xp =
+        if is_integer(total.total_xp) do
+          total.total_xp
+        else
+          decimal_to_integer(total.total_xp)
+        end
```
RichDom2185, line 1486: If it's not an integer, what will it be?

RichDom2185 (changes requested):

lib/cadet/assessments/assessments.ex
thortol: This check was added because the code was failing test cases. In the test cases, total.total_xp is seen as nil which causes an error to occur.

RichDom2185: But what does it mean, logically? Why would you expect total XP to be nil?

thortol: Based on the current workflow it will not be nil as when a value is inserted it will be set to 0. However, as xp does not have a not null constraint, it is still possible to be nil so having this redundant check will still be useful. Or would it be better to fix the test cases and set xp to 0?

PR #1180, opened 2024-09-08
Transfer groundControl (and admin panel) from staff to admin route

josh1248:
Accompanies the frontend pull request at https://github.com/source-academy/frontend/pull/3043

Previously, "admin" routes were available to both staff and admins to access.

Now, a new "staff" route has been created that accepts most of the previous functions. The remaining functions in "admin" scope have the ability to cause massive damage, such as the deletion of entire assessments. Hence, they are now inaccessible to avengers even if they bypass the frontend route checking.

Aside: I have also shifted admin panel actions into this admin scope.


josh1248:
oh no :(

josh1248:
The new scope has caused the test error codes to change from 400 to 403 - will update the tests for staffs and admins.

josh1248:
image

Error messages are no longer being generated at the controller level but rather at the router level due to the changes, which causes the tailored error messages to be hidden. I was wondering if this was something I should revert? Alternatively, it could be possible to make the admin check a per-route check, allowing for different error messages rather than a generic "403 Forbidden". Would take some time, though (should still be manageable to bulk-change).

RichDom2185 (changes requested):
Can you revert all the unnecessary changes? Refer to the diff for the (I think mostly autogenerated) changes. Thanks!

josh1248:
Apologies for missing this out - I have reverted the accidental formatting changes!

RichDom2185 (approved):
LGTM, thanks!

PR #1188, opened 2024-09-19
fix: Fix autopublish grading by auto grader

GabrielCWT:
If answers were sent to the AWS Lambda for autograding, i.e. Programming questions, when updating the answer, it does not check if the assessment requires manual grading. This results in the submission being autopublished


RichDom2185 (approved):
LGTM, thanks! Sorry for the delayed review&hellip;

PR #1190, opened 2024-09-22
Fix: Fix limit function for grading table fetch function 

GabrielCWT:
- Changes atom to use correct atom (`page_size` instead of `limit`)
- Implement tests for pagination to prevent this from happening again


RichDom2185 (approved):
LGTM, thanks!

PR #1200, opened 2024-10-13
fix: Auto submission cron job for team assessments bug

### LINKED ISSUE #1199: Auto submission cron job creates empty individual submissions despite having a team submission
As per title.

For team assessments, despite the student already having a submission for their team, when the deadline comes, the auto-close/auto-submit cron job is not able to find their submissions, and as a result, two things happen:

* Their actual team submission is not autoclosed and submitted, potentially allowing them to bypass the deadline
* A new empty individual submission is created for each student, resulting in a precondition error when they try to load their assessments page (since there are 2 submissions associated with the same assessment for the same student – their individual one and team one)

RichDom2185:
Closes #1199.

### DIFF
--- a/lib/cadet/accounts/teams.ex
+++ b/lib/cadet/accounts/teams.ex
@@ -9,8 +9,8 @@ defmodule Cadet.Accounts.Teams do
   import Ecto.{Changeset, Query}
 
   alias Cadet.Repo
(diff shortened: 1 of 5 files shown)


GabrielCWT (approved):
LGTM

PR #1210, opened 2024-11-13
Create endpoint for retrieval of all assessment submissions

josh1248:
Frontend PR: https://github.com/source-academy/frontend/pull/3076
Branched from #1180

Creates an endpoint that obtains all assessments submitted without any filter or pagination. Admin-only endpoint.

### DIFF
--- a/lib/cadet_web/admin_controllers/admin_grading_controller.ex
+++ b/lib/cadet_web/admin_controllers/admin_grading_controller.ex
@@ -60,17 +60,15 @@ defmodule CadetWeb.AdminGradingController do
 
   @doc """
   # Description
-  Attains every assessment submission registered, regardless of grading or attempted status.
-  This is a special call of the general `index` submission whereby pagination is bypassed.
+  Attains every assessment submission registered, regardless of grading
+  or attempted status. This is a special call of the general `index`
+  submission whereby pagination is bypassed.
   """
   def index_all_submissions(conn, _) do
     # 100 billion page size here should cover all feasible table rows in 1 page
     # given that the hard limit of pages in a postgres database is ~4 billion.
     # (https://www.postgresql.org/docs/current/limits.html)
-    index(
-      conn,
-      %{"group" => "false", "pageSize" => "100000000000", "offset" => "0"}
-    )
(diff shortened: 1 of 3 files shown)


lib/cadet_web/admin_controllers/admin_grading_controller.ex
```diff
@@ -58,6 +58,17 @@ defmodule CadetWeb.AdminGradingController do
     index(conn, %{"group" => "false"})
   end
 
+  def index_all_submissions(conn, _) do
+    index(
+      conn,
+      %{
+        "group" => "false",
+        "pageSize" => "100000000000",
```
RichDom2185, line 66: Haha sure, maybe in the future we can filter by assessment name (it's a use case for the PA because we export the grades to Canvas separately).

priv/repo/seeds.exs
```diff
@@ -26,7 +26,7 @@ alias Cadet.Accounts.{
 # Cadet.Repo.insert!(%Cadet.Settings.Sublanguage{chapter: 1, variant: "default"})
 
 if Cadet.Env.env() == :dev do
-  number_of_students = 100_000
+  number_of_students = 100
```
RichDom2185: Any reason this is reduced again?

RichDom2185 (approved):
LGTM, thanks

priv/repo/seeds.exs
josh1248: it was taking too long to reset the db :(

lib/cadet_web/admin_controllers/admin_grading_controller.ex
josh1248: oh yes, that would be a good future change to generalize the export function!

priv/repo/seeds.exs
RichDom2185: Oof 😢 ok maybe we will see how then, we can change it back later haha

RichDom2185: Though I think we can put it to 1000 to mimic 1101S?

josh1248: yeah, 1000 would be good! doesnt take too long for a one time set up (at most a minute).

PR #1236, opened 2025-03-04
Added Exam mode 

iZUMi-kyouka:
### Description

- Added exam mode functionality in the backend by adding enable_exam_mode and is_official_course in the courses table.
- Added get_exam_mode_course() function in lib/cadet/accounts/course_registrations.ex
- Modified lib/cadet_web/controllers/user_controller.ex logic in serving the last_viewed_course to always serve the official institution course under exam_mode if any of the user's official course is currently under exam mode. Performed the same modification for get_latest_viewed() function in the same file.
- Added the required migrations file to alter the database
- Added the 2 additional fields in the model and wherever necessary

01/04/25
- Added resume_code to courses table
- Added isPaused to users table
- Added some tests for enable_exam_mode, is_official_course, and resume_code for the changesets

09/04/25
- Added user_browser_focus_log table in the DB
- Added the required endpoint to log when a user's browser loses focus and regains focus when under exam mode

Note: this may require change to the DB diagram in README.md

### Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update
- [ ]  Code quality improvements

### Checklist

- [x] I have tested this code


RichDom2185 (commented):
We don't need `isOfficialCourse` anymore, right?

RichDom2185 (commented):
Please fix the failing tests, thanks!

lib/cadet_web/controllers/user_controller.ex
```diff
@@ -12,45 +12,71 @@ defmodule CadetWeb.UserController do
...
-        latest: nil,
-        max_xp: nil,
-        story: nil,
-        xp: nil
-      )
+    exam_mode_course = CourseRegistrations.get_exam_mode_course(conn.assigns.current_user)
+
+    cond do
+      exam_mode_course ->
+        IO.puts("Course #{exam_mode_course.course_id} is under exam mode.")
```
GabrielCWT, line 19: Probably should remove this.

lib/cadet_web/controllers/user_controller.ex
```diff
@@ -12,45 +12,71 @@ defmodule CadetWeb.UserController do
...
-      end
+    if exam_mode_course do
+      latest = CourseRegistrations.get_user_course(user.id, exam_mode_course.course_id)
+      get_course_reg_config(conn, latest)
+    else
+      latest =
+        case user.latest_viewed_course_id do
+          nil -> nil
+          _ -> CourseRegistrations.get_user_course(user.id, user.latest_viewed_course_id)
+        end
```
GabrielCWT, line 76: Could we make this more concise? Something like
```
latest = cond do
  exam_mode_course -> CourseRegistrations.get_user_course(user.id, exam_mode_course.course_id)
  user.latest_viewed_course_id -> CourseRegistrations.get_user_course(user.id, user.latest_viewed_course_id)
  true -> nil
end

get_course_reg_config(conn, latest)
```

lib/cadet/courses/course.ex
```diff
@@ -28,6 +31,9 @@ defmodule Cadet.Courses.Course do
     field(:enable_achievements, :boolean, default: true)
     field(:enable_sourcecast, :boolean, default: true)
     field(:enable_stories, :boolean, default: false)
+    field(:enable_exam_mode, :boolean, default: false)
+    field(:resume_code, :string)
+    field(:is_official_course, :boolean, default: false)
```
GabrielCWT, line 36: As Richard pointed out, is this needed?

GabrielCWT (changes requested):

lib/cadet/courses/course.ex
iZUMi-kyouka: Yes, these are needed for the additional functionality we require.

lib/cadet_web/controllers/user_controller.ex
iZUMi-kyouka: Noted, thanks for the feedback.

lib/cadet_web/controllers/user_controller.ex
iZUMi-kyouka: Oops yes, it was used for debugging, thanks!

lib/cadet_web/controllers/courses_controller.ex
```diff
@@ -40,6 +44,42 @@ defmodule CadetWeb.CoursesController do
     end
   end
 
+  def check_resume_code(conn, %{"course_id" => course_id}) when is_ecto_id(course_id) do
```
GabrielCWT, line 47: This function is doing more than just checking resume code, possible to refactor some logic out into its own function? i.e. updating the user's `is_paused`

lib/cadet/accounts/user.ex
```diff
@@ -21,6 +21,7 @@ defmodule Cadet.Accounts.User do
     field(:provider, :string)
     field(:super_admin, :boolean)
     field(:email, :string)
+    field(:is_paused, :boolean)
```
GabrielCWT, line 24: Is there a reason why this is linked to the user and not their course_registration? Isn't exam mode related to their course.

lib/cadet_web/controllers/user_controller.ex
```diff
@@ -12,42 +12,68 @@ defmodule CadetWeb.UserController do
...
+      exam_mode_course ->
+        xp = Assessments.assessments_total_xp(exam_mode_course)
+        max_xp = Assessments.user_max_xp(exam_mode_course)
+        story = Assessments.user_current_story(exam_mode_course)
+
+        render(
+          conn,
+          "index.json",
+          user: user,
+          courses: courses |> Enum.filter(fn c -> c.course_id == exam_mode_course.course_id end),
```
GabrielCWT, line 27: Is there a difference between `exam_mode_course` gotten? Why not just have courses contain `exam_mode_course` which you have already obtained.

iZUMi-kyouka: The filter is so that the list of courses returned, if any of the user's course is under exam mode, only contain this particular course that is under exam mode, so the what the frontend sees will only be this course.

lib/cadet/accounts/user.ex
iZUMi-kyouka: Initially I thought linking this to the user would allow for the enforcing of the pause beyond the course so that if a user is paused due to opening other app / using dev tool (which is our plan), the user will have to settle the problem with the course admin / coordinator. But, now that you pointed out, maybe this should not affect the user through all their courses. Should I move this course_registration?

lib/cadet_web/controllers/courses_controller.ex
iZUMi-kyouka: Noted!

lib/cadet_web/controllers/user_controller.ex
GabrielCWT: I understand, so wouldnt that be the exam course you have already fetched? Is there a need for this filter.

iZUMi-kyouka: AH so sorry I realise what you initially meant! Yes I'll just return the exam course that is already fetched! Thanks!

GabrielCWT:
One more thing which just came to mind. Correct me if I am wrong, your current implementation places all users with a course which has exam mode enabled have their SA placed on some sort of "lock down". Wouldn't this limit the use for admins/staff during the time in which the course is on lock down?

iZUMi-kyouka:
> One more thing which just came to mind. Correct me if I am wrong, your current implementation places all users with a course which has exam mode enabled have their SA placed on some sort of "lock down". Wouldn't this limit the use for admins/staff during the time in which the course is on lock down?

Yes this is correct as of now. I will work on excluding them, so that staffs and admins work as usual, thanks!

PR #1239, opened 2025-03-19
Add on_finish field to assessments

JustATin555:
- Added `is_minigame` string field to `AssessmentsConfig`
- Added `isMinigame` field to JSON api

Requires all users to migrate with `mix ecto.migrate`
Should be used in conjunction with https://github.com/source-academy/frontend/pull/3105


GabrielCWT:
Discussed with @JustATin555, this should be moved to assessment config instead of in assessment

lib/cadet/courses/assessment_config.ex
```diff
@@ -18,6 +18,7 @@ defmodule Cadet.Courses.AssessmentConfig do
     field(:early_submission_xp, :integer, default: 0)
     field(:hours_before_early_xp_decay, :integer, default: 0)
     field(:is_grading_auto_published, :boolean, default: false)
+    field(:is_minigame, :boolean, default: false)
```
GabrielCWT, line 21: Could you add a comment on what it is used for.

GabrielCWT (changes requested):

lib/cadet/courses/assessment_config.ex
JustATin555: Added

GabrielCWT (approved):

RichDom2185 (approved):
LGTM, thanks! Will merge since it's not a blocker for the frontend PR

PR #1284, opened 2025-08-14
fix: Create log stream if it doesn't exist

RichDom2185:
So that when the instance starts up (and the unix time changes)/with multiple instances on different nodes, it can automatically create the new stream


tckeong (approved):
LGTM! I hadn’t considered that scenario—thanks for improving it.

PR #1288, opened 2025-08-19
Leaderboard fixes and improvements

sayomaki:
Small PR to make some fixes and changes to the leaderboard backend logic.

- Update and fix submissions XP query
- Use pipeline operations instead of direct queries
- Make query options as parameter map to allow more options


RichDom2185 (approved):
LGTM, thanks! Didn't have time to test but looks good

PR #1340, opened 2026-03-01
Support NUS Entra ID as authentication provider

RichDom2185:
As per title. Also fixed tool version.


sayomaki (approved):
Not tested. Hope this works!

PR #1364, opened 2026-07-25
Add assessment level language and evaluator config fields

Akshay-2007-1:
## Summary

Adds nullable language_id and evaluator_id columns on assessments so that a mission, quest, path or contest can be configured with a Conductor backed language (e.g. Python), instead of relying only on the global Playground language selector. This mirrors the frontend WIP in source-academy/frontend#4057, which handles the Run and testcase grading paths for Conductor languages but currently has no way to know which language a given assessment should use.

Scope, following the direction agreed with Prof Henz: assessment level config only, not per question. The existing Library.chapter/variant fields on questions are untouched.

## Changes

- New migration adding language_id and evaluator_id (both nullable strings, default nil) to the assessments table. Existing assessments are unaffected and keep current js-slang behaviour until a value is set.
- Assessment schema and changeset updated to accept the two new optional fields.
- Admin update endpoint accepts languageId and evaluatorId in the update payload, following the same pattern as existing optional fields like hasTokenCounter.
- Both the student facing and admin facing assessment views serialise the new fields in the overview and show responses.
- Updated existing controller tests that assert exact response bodies to include the new keys.

## Test plan

- [x] mix format
- [x] mix compile with no new warnings
- [x] mix test (assessment related suites run individually, then full suite via pre push hook)
- [ ] Manual verification once the frontend side is wired up to actually set and read these fields for a Python assessment

### DIFF
--- a/lib/cadet/assessments/assessment.ex
+++ b/lib/cadet/assessments/assessment.ex
@@ -38,6 +38,8 @@ defmodule Cadet.Assessments.Assessment do
     field(:has_voting_features, :boolean, default: false)
     field(:llm_assessment_prompt, :string, default: nil)
     field(:is_autosave_enabled, :boolean, default: true)
+    field(:language_id, :string, default: nil)
+    field(:evaluator_id, :string, default: nil)
 
     belongs_to(:config, AssessmentConfig)
     belongs_to(:course, Course)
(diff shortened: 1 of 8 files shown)


Akshay-2007-1:
Closing — mystery solved in #1366: DEPLOYMENT was renamed to PROGRAMMINGLANGUAGE in PR #1247 (Nov 2025), backend already handles this correctly under the new name. No backend schema/parsing changes needed after all, we just need to rename the tags in the XML content itself.

PR #1370, opened 2026-07-30
Add hardened mode restricting course creation to super admins

RichDom2185:
## Summary

Adds a config-driven flag that restricts course creation (`POST /v2/config/create`) to global super admins.

When enabled, any non-super-admin attempting to create a course receives an **opaque `403 "Forbidden"`** — the same generic body the router's authz plug returns — so the restriction is not disclosed to the client. The real reason is logged server-side only. The flag defaults to **off**, so existing behaviour is unchanged unless explicitly enabled. **No frontend changes required.**

## Configuration

Opt-in via `config/` files:

```elixir
config :cadet, restrict_course_creation: true
```

Read through the new `Cadet.Env.restrict_course_creation?/0` helper; documented for prod runtime in `config/cadet.exs.example`.

## Changes

- `config/config.exs` — declare `restrict_course_creation` flag (default `false`)
- `lib/cadet/env.ex` — add `Cadet.Env.restrict_course_creation?/0`
- `lib/cadet_web/controllers/courses_controller.ex` — enforce in `create/2`
- `config/cadet.exs.example` — document the prod runtime option
- `test/cadet_web/controllers/courses_controller_test.exs` — cover super-admin success and non-super-admin opaque 403

## Testing

- `mix test test/cadet_web/controllers/courses_controller_test.exs` → **11 tests, 0 failures** (9 existing + 2 new)
- `mix format --check-formatted` → clean
- `mix credo` → no issues on changed source

## Notes

Rejections return **403 Forbidden** with a generic, opaque body — consistent with the sibling too-many-courses branch and the router's role-based rejections. (An earlier revision used 401, but that would incorrectly trigger the frontend's token-refresh flow.)

### DIFF
--- a/config/cadet.exs.example
+++ b/config/cadet.exs.example
@@ -152,6 +152,9 @@ config :cadet,
 # config :cadet, Cadet.Jobs.Scheduler,
 #   timezone: "Asia/Singapore",
 
+# Hardened mode: when true, only super admins may create courses.
+# config :cadet, hardened_mode: true
+
 # # Additional configuration for SAML authentication
 # # For more details, see https://github.com/handnot2/samly
 # config :samly, Samly.Provider,
--- a/config/cadet.exs.example
+++ b/config/cadet.exs.example
@@ -152,8 +152,8 @@ config :cadet,
 # config :cadet, Cadet.Jobs.Scheduler,
 #   timezone: "Asia/Singapore",
 
-# Hardened mode: when true, only super admins may create courses.
-# config :cadet, hardened_mode: true
+# When true, only super admins may create courses.
+# config :cadet, restrict_course_creation: true
 
 # # Additional configuration for SAML authentication
 # # For more details, see https://github.com/handnot2/samly
--- a/config/config.exs
+++ b/config/config.exs
@@ -11,6 +11,9 @@ config :cadet, environment: Mix.env()
 config :cadet,
   ecto_repos: [Cadet.Repo]
 
+# Hardened mode: when true, only super admins may create courses.
+config :cadet, hardened_mode: false
+
 config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
 
 # Scheduler, e.g. for CS1101S
--- a/config/config.exs
+++ b/config/config.exs
@@ -11,8 +11,8 @@ config :cadet, environment: Mix.env()
 config :cadet,
   ecto_repos: [Cadet.Repo]
 
-# Hardened mode: when true, only super admins may create courses.
-config :cadet, hardened_mode: false
+# When true, only super admins may create courses.
+config :cadet, restrict_course_creation: false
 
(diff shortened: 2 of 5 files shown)


sayomaki:
I'm unsure about whether Elixir's strict boolean semantics would be a problem here: `and` and `or` require boolean left operands and raise `BadBooleanError` if given anything else, while `not` requires a boolean and raises `ArgumentError` otherwise; this is distinct from the truthy-based &&, ||, and ! operators.

sayomaki:
## 1. The `get_env/3` fallback is dead code today

`config/config.exs` is evaluated for every Mix env and baked into the release's `sys.config` at build time. The key is therefore *always* present at runtime, so the third argument to `Application.get_env/3` can never fire.

It's defensive code that no execution path reaches — which means it will never be exercised, and nothing will notice if it drifts out of sync with the value above it.

## 2. The moment it stops being dead, it disagrees

If the declared default is ever flipped — `config :cadet, restrict_course_creation: true` — the two defaults now say opposite things, and the fallback governs precisely the case where the compiled config is absent. It says **permissive**.

For a security control, "unconfigured" resolving to "unrestricted" is the wrong direction. A release built without `config.exs` applied, or anything calling `Application.delete_env/2`, silently disables the restriction rather than refusing to start. The failure is invisible: no log line, no boot error — courses just quietly become creatable again.

## 3. Neither file is marked authoritative

`config/cadet.exs.example` documents *how to turn the flag on* but says nothing about where the default is declared. Two maintainers asked to "change the default" will edit two different files, and each will be half-right.

## 4. No normalisation — a malformed config value 500s the endpoint for everyone

`@spec restrict_course_creation? :: boolean()` is a promise, not an enforcement. Dialyzer cannot check it, because `Application.get_env/3` is typed `any()`. Whatever is in `/etc/cadet.exs` flows through unvalidated to the call site:

```elixir
Cadet.Env.restrict_course_creation?() and not user.super_admin
```

`and` / `or` / `not` are Elixir's **strict** boolean operators — they require actual booleans, unlike `&&` / `||` / `!` which accept any truthy value.

`/etc/cadet.exs` is a hand-maintained file, and operators wiring config from environment variables produce strings. So `restrict_course_creation: "true"` is a plausible typo, and it raises `BadBooleanError` before the right operand is evaluated (`and` short-circuits only on `false`, so a non-boolean on the left always raises).

**Failure scenario.** An operator enables the restriction with a stringly-typed value. Every `POST /v2/config/create` now returns **500 — including super admins'**. The intended-privileged path breaks along with the restricted one. The exception message ("expected a boolean on the left side of `and`") points at `courses_controller.ex` rather than at the config file, so the operator has no reason to suspect their config edit.

Note that `"false"` fails identically — this isn't truthiness biting, it's that *any* non-boolean raises.

sayomaki (approved):
LGTM

PR #1377, opened 2026-08-01
fix(courses): correct pattern match on create_course_config result

RichDom2185:
## Problem

Course creation crashes with a 500 despite the course being created successfully. Sentry:

> `KeyError: key :id not found` at `lib/cadet_web/controllers/courses_controller.ex:54`

## Root cause

`Courses.create_course_config/2` returns the raw `Ecto.Multi` transaction result, which on success is:

```elixir
{:ok, %{course: %Course{...}, course_reg: %CourseRegistration{...}}}
```

The controller matched `{:ok, course}`, binding the **whole changes map** to `course`. The success-logging line then evaluated `course.id`, which raised `KeyError` because the map has `:course`/`:course_reg` keys, not `:id`. The transaction had already committed, so the course was created but the client received a 500 instead of `"OK"`.

The context function itself already destructures this shape correctly in its own logging (`courses.ex:45`), confirming the expected shape.

## Fix

Match `{:ok, %{course: course}}` in the controller to extract the actual `Course` struct before logging `course.id`.

## Regression

Introduced in #1286, which added the `Logger.info("... #{course.id} ...")` line to the previously value-ignoring `{:ok, _} ->` branch. #1370 reindented the block but carried the bug forward.

## Notes

- The existing "succeeds" controller test (`test/cadet_web/controllers/courses_controller_test.exs:19`) exercises this line and asserts a 200, so it should have been failing on `master` — worth checking why CI didn't catch this.

🤖 Generated with [[Claude Code](https://claude.com/claude-code)](https://claude.com/claude-code)

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -8,8 +8,8 @@ config :cadet, CadetWeb.Endpoint,
 
 config :cadet, environment: :test
 
-# Print only warnings and errors during test
-config :logger, level: :warning, compile_time_purge_matching: [[level_lower_than: :warning]]
+# Run at :info so interpolated log expressions are evaluated (and thus tested)
+config :logger, level: :info
 
 config :ex_aws,
   access_key_id: "hello",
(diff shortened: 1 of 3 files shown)


sayomaki (approved):
LGTM

PR #1381, opened 2026-08-06
Fix 422 on grading save when no AI comment log exists

### LINKED ISSUE #4233: Grading tab: save changes gives 422
https://github.com/user-attachments/assets/82b4c034-d5cb-4e72-9286-2114de6fbf3c

martin-henz:
## Summary
- Saving grading changes from the Grading tab (frontend) always fails with a 422 "Error while communicating with backend" toast, even though the actual grade/XP save succeeds separately.
- Root cause: the frontend calls `POST /admin/save-final-comment/:answer_id` on every grading save (used to persist the grader's final comment against the AI-comment log, for the AI-assisted commenting feature). `Cadet.AIComments.update_final_comment/2` returned `{:error, :not_found}` whenever the answer had no AI-comment log row — which is true for the overwhelming majority of submissions, since AI-assisted commenting is opt-in. The controller only distinguished `{:ok, _}` from `{:error, changeset}`, so `:not_found` fell into the generic changeset-failure branch and returned 422.
- Fix: treat a missing AI comment log as a no-op success (`{:ok, nil}`) instead of an error, since there is nothing to attach the final comment to in that case.

Fixes source-academy/frontend#4233

## Test plan
- [ ] Confirmed locally via source reading + live reproduction on staging: `POST /admin/save-final-comment/:answer_id` returned 422 for a submission whose answer had no prior AI-comment log, while the parallel `POST /admin/grading/:submissionId/:questionId` returned 200 for the same click.
- [ ] No local Elixir/`mix` toolchain was available in this environment to run `mix test` or `mix compile --warnings-as-errors` — please verify CI passes.
- [ ] Manually verify on staging/review app: open the Grading tab for a submission that has never had AI comments generated, edit the comment/XP, click "Save Changes", and confirm no 422 toast appears.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011Lfqxf3TfRtt4PLB92u11S


sayomaki (approved):
LGTM

